PR #2, opened 2016-01-16
installer script creates ear for appegine

mscurtescu:
This is the new single commit which uses my @google.com email address.


mitchellsundt:
I pulled this into the public gear branch (master points to the stable release tip).

PR #5, opened 2017-01-26
In CONFIGURE.txt change security.settings to security.properties

kden:
Also trying out a pull request.


kden:
There is a mismatch between what's in odk-gae-settings and what's in CONFIGURE.txt, but this pull request doesn't accurately fix it.  Please ignore.

PR #6, opened 2017-04-14
Docs re-org

daveycrockett:
I just walked through the dev environment spin-up process and thought I'd take a crack at reorganizing the documentation to make it a little more user-friendly for viewing within git.  Changed CONFIGURE.txt to markdown, split it across multiple files.  Only downside is this leads to a commit diff that isn't particularly easy to compare content.  Open to suggestions on how to improve that.


daveycrockett:
Closing to PR to the appropriate branch.

PR #110, opened 2017-10-09
Remove duplicate lists of classes used for JavaRosa serialization

dcbriccetti:
Use classnames newly exposed from JavaRosa, rather than maintaining this list of classnames here as well as in JavaRosa (and in Collect).
Depends on https://github.com/dcbriccetti/javarosa/commit/15a825805a3f5976a908d0e8a1b65c7f8642cf96 .

@mitchellsundt, how does this look?

I can’t yet build and test Aggregate. I wonder if one of you who can could test this?

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
+++ b/src/main/java/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
@@ -29,6 +29,7 @@ import java.util.Map;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.javarosa.core.model.CoreModelModule;
 import org.javarosa.core.model.DataBinding;
 import org.javarosa.core.model.FormDef;
 import org.javarosa.core.model.IDataReference;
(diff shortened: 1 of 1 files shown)


lognaturel:
Let's come back to this one after JR 2.6.0 is released and integrated into Aggregate and then #112 is merged. At that point you should be able to easily run and verify it, @dcbriccetti (and if not we can file issues to further improve the build process).

@ggalmazor does this sound ok to you?

ggalmazor:
Sounds good. I've already checked that it merges after #112 without any problem.

ggalmazor:
@dcbriccetti, I'm testing this PR... How could I check that these changes work? I mean, the code compiles and all, but, is there some check I could do to be sure? The answer can be "no", that's fine :)

yanokwa:
@lognaturel I haven't been able to verify this change, but since it's a pain to verify without the changes that #122 provides, I propose we merge it. It's relatively low risk because 
1. Dave is generally very careful.
1. Guillermo has tested it.
1. If this fails, we'll see it way before we do a release.

lognaturel:
This is a change we've made in the other tools too and as @ggalmazor we can visually verify that the new JR API registers the same classes as before.

Issue #102, opened 2017-09-07
Store attached CSVs, link to them if available


lognaturel:
opendatakit/xforms-spec#94 describes a standard for specifying that a client should generate an audit trace for a specific form and for specifying the audit csv filename in the submission.

In the Aggregate table view, the user should be able to click on the audit filename to see a popup with the audit contents the same way that he/she can click on an image to get a popup with the image.

A form for testing: [audit-test-orx.xml.txt](https://github.com/opendatakit/aggregate/files/1285353/audit-test-orx.xml.txt)

dcbriccetti:
A simpler form:

[shorter-audit-test-orx.xml.txt](https://github.com/opendatakit/aggregate/files/1364277/shorter-audit-test-orx.xml.txt)

ggalmazor:
Can this issue be closed? @yanokwa @lognaturel

lognaturel:
Absolutely. Looks like "solves" is not a [keyword for closing issues](https://help.github.com/articles/closing-issues-using-keywords/) so that's why it didn't close automatically!

PR #111, opened 2017-10-13
New feature that lets a user see Audit CSV contents inside a popup in Aggregate's submissions table

ggalmazor:
This PR solves #102 

It covers just a basic feature to enable users to see Audit CSV contents inside a simple popup. 

Some UI improvements that could be addressed:
 - New button image (currently using preview button image)
 - Some styling or HTML structuring to the Audit CSV data (maybe formatting it into a table)

Some code improvements that could also be addressed:
 - Reduce duplication of code in `SubmissionServiceImpl` and `BinaryDataServlet`

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/client/popups/AuditCSVPopup.java
+++ b/src/main/java/org/opendatakit/aggregate/client/popups/AuditCSVPopup.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright (C) 2011 University of Washington
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


lognaturel:
Verified with both `orx:audit` (most strictly correct) and `audit` (acceptable). Also verified that Briefcase pulls down the audit csv.

Line breaks don't seem to be respected so it's really hard to read! I think it's worth putting into a table now so it's more usable. I'm also surprised to see that the text can be selected but not copied. It would be useful for it to be able to be copied.

ggalmazor:
> "the text can be selected but not copied"
That's some GWT black magic there :/ I'll look into that!

ggalmazor:
@lognaturel could we merge this one and wait for #112 to be merged before improving the Audit CSV popup? Fixing this on master would require me to set up the old Eclipse dev environment and I could go faster on the new Gradle project...

lognaturel:
Merged, @ggalmazor! That does make more sense.

Issue #106, opened 2017-09-12
Switch to Gradle?


brmeyer:
If I'm honest, I'm not a huge Gradle fan, at least not in most contexts.  But ODK Aggregate is one example where I think it may truly help.

For projects that simply need standard dependency management, standard composability, deployment/release management, and lightweight script/plugin use, Maven is still solid.  But where it Gradle shines is its ability to use Groovy scripts for dynamic and complex needs, such as Aggregate's GAE setup.  Rather than rely on a combo of Maven, Maven plugins (ugh), and Ant scripts, Gradle/Groovy has a pretty big advantage.

Other benefits:

- IDE support has come a **long** way in the last couple of years.
- Consistency with ODK's Android projects.
- Potential to replace the installer setup in its entirety, not requiring Maven to be first installed, etc.  Let the Gradle wrapper do its thing, keeping the entire setup self-contained and headless.  The end result would be the creation of deployable WAR, or completely scripting and automating a PaaS deployment.

Worries:

- It's improved lately, but Gradle can have slower build times than Maven.  But with the module separation suggested in https://github.com/opendatakit/aggregate/issues/105, this can mostly be mitigated for developers.
- Gradle scripts can be tricky to debug -- there seem to be 100 ways to shoot yourself in the foot.  But, arguably, Maven + Ant can be worse...

Curious to hear what everyone else things!  Maven vs Gradle vs whatever Google happens to be pushing seems to be rather divisive ;)

lognaturel:
@arahayrabedian @kden @jovaee Are some of the people I can think of who were compiling Aggregate recently and might have some feedback.

I personally find the benefits compelling and the worries not all that worrisome so I'd be on board! I particularly like the idea that we'd have fewer tools to know across the different ODK projects.

yanokwa:
@johnhbenetech It'd be good to get your feedback here too!

kden:
So are we talking about continuing on Aggregate + ODK 1.0?  Will Aggregate continue to be relevant in 2.0?  Aggregate supports a lot of different technologies; if it continues on, I think it should be spun off into a GAE project, a relational DB project, and a shared code project.  Gradle is a natural for the GAE project; for a Tomcat/RDB project, it's more of a toss-up.

lognaturel:
[...]

Issue #105, opened 2017-09-12
Re-architect the modules with clear separation between service/business src and infrastructure


brmeyer:
Morning!  I'm hoping to help think through what it would look like to re-architect the project setup, with a primary goal: Make it as easy as possible for new and existing contributors to be effective.  The following is purely a WIP that I'll keep adding to as I go through the project.  These are purely suggestions based on approaches I've used in past lives, but I'd love feedback from others!

Some inspiration is drawn from https://github.com/ArtificerRepo/artificer, which is an artifact and metadata repository I used to be involved with @ Red Hat.  That project has a lot of warts and I'm definitely not claiming it to be a perfect role model.  But we ran into similar issues: various src layers, concurrent support for numerous environments/containers (some of which require substantial config and artifact changes), integration test harnesses, tooling, clients, etc., all while remaining IDE-agnostic and attempting to make new contributions as easy as possible.  Success is arguable ;), but there were some steps in the right direction.

- /aggregate
  - /aggregate-src
    - _`aggregate-src` currently exists, but it's only a POM that refers to the `src` dir back at the top level.  This pulls together all service, business, etc. src -- likely most non-test and non-infra code -- into a jar that can be used as a dependency by infra modules._
    - _Note that aggregate-src could/should be broken down into additional modules, each its on jar and usable as a dependency.  As is, the code base is fairly large.  Plus, this is helpful for composibility, if separation of concerns is currently possible._
    - Artifact: jar
    - /src (main and unit test)
  - /server
    - /api
      - _Houses any Servlet classes, controllers, or any "API" src that's specific to the generic application server._
      - Dependencies: aggregate-src, servlet-api
      - Artifact: jar
    - all Tomcat, GAE, etc. specific submodules
      - _Isolate all container-specific configuration, dependencies, and dynamic tasks (Ant or otherwise)._
      - Dependencies: aggregate-src, api, container-specific dependencies
      - Artifact: specific to the container (war, ear, etc.)
    - /dev-server
      - _Provides a harness that's primarily for developers, giving a simple way to kickoff an embedded server (Tomcat/Jetty/Undertow) that automatically pulls in necessary deps and jars/wars.  Likely resembles the other container-specific submodules, but super stripped down.  This may be what some of the Eclipse-specific bits are doing now, but in an IDE-agnostic way.  Easily used from any IDE run config._
  - /integration-test
    - _Isolate all integration test projects here (most are currently in `/aggregate/src/it`)._
[...]

PR #112, opened 2017-10-15
Project migration to Gradle

ggalmazor:
This PR consists on a complete restructuration of the project to use Gradle as a build tool.

As a result, Eclipse support is dropped in favor of IntelliJ integration.

The documentation has also been updated matching the structure from other ODK projects like ODK Collect.

This PR is based on the feedback and work done by other contributors, especially #105 and #106 among others.

This PR is not proposed as a final Gradle/IntelliJ integration solution but as a step in the same direction other ODK projects have taken. I'm sure I'm introducing problems which are already covered by the current Eclipse workflow and there is lots of room for improvements upon the proposed changes that we need to work on. 

**Note:** This PR pulls #111 along

### Warning

This is a breaking change that affects the current development workflow and build process. There isn't an easy migration path. Therefore, starting from scratch and creating a new project following the new instructions is encouraged to propertly test this PR.

### What has been done?

The first goal was to transition to Gradle in a single project structure

- Copied `src/main/java` into an empty Gradle project
- Copied a fixed set of configuration files for a Tomcat + PostgreSQL setup to `src/main/resources`
- Worked out the dependencies from the original `pom.xml` file
- Used GWT plugin to add a GWT compilation phase
- Configured WAR plugin to generate the same output as the original Eclipse build process. Exploded WAR files should have the same contents.
- Created and tested a `Run/Debug configuration` in IntelliJ that would run Tomcat 8.0 and deploy the WAR generated by Gradle. This configuration would have to let you debug server code.

Once this was achieved, the second goal was to separate a common Gradle submodule from specific tomcat & gae submodules.

- Created Gradle submodules
- Moved all the code to `common` module
- Extracted any `gae` or `tomcat` related code to their respective submodules
- Diff'ed the configuration files for gae and tomcat variants from the original project to get a sense of what were the difference and created versions of them in `src/main/resources` in their respective submodules
- Applied same process to `webapps` related stuff: `web.xml`, gae specific stuff, etc.
- Configured the WAR plugin on `tomcat` and `gae` submodules to depend on `common` WAR task and to compose its contents

After this, you can build the main project and obtain a WAR file both for gae and tomcat or just build the submodule you need with `gradle :tomcat:build` for example.

The third goal was to bring all the testing code and the BitRock installer stuff.

- Moved original `src/test` source set into `common` module's test source set
- Distributed specific tomcat or gae tests to their respecting submodules
- Created a new `functionalTests` Gradle submodule for original `src/it` test source set. The goal here was to remove selenium and functional testing stuff from the main project and to create an opportunity to reflect on the current testing strategy.
- Didn't bring performance tests from `src/itperf`. There was no information on how to run them and while trying to reverse engineer them, apparently you need an access token from Google's OAuth v2 platform.
- Copied and adapted original BitRock installer stuff

### Stuff that could/should be improved

- Support for GWT dev/superdev modes
- Rethink testing strategy: create unit tests for key core features, create integration tests for untested parts of database and web interactions, apply snapshot testing/approval testing to client's UI, etc.
- Should different test source sets be needed (for unit and integration tests, for example) inside the same Gradle submodule, follow [Petri Kainulainen's guide](https://www.petrikainulainen.net/programming/gradle/getting-started-with-gradle-integration-testing/)
- Preparing stuff for BitRock installer could be simplified by replacing it for a shell script
- Use Gretty plugin to run `common` module on an embedded Jetty instance for development

### Possible problems that should have to be checked

Eventhough I have checked most of the following items, it would be great if the community would help checking them as well.

- Lost/buggy features in Aggregate due to new WAR packaging process.
- Google App Engine support
- Missing runtime dependencies
- Wrong runtime jdbc driver dependencies
- Effect of losing transitive package exclusions
- Is BitRock installer working as intended? I think there might be some problems on the configuration files while installing

Depending on your experience with Aggregate you could even find other things that I've not though about. 

Please, send your feedback ;)


lognaturel:
@ggalmazor thank you so much for the detailed writeup and the super easy process. This really nicely matches build processes for other tools.

@wbrunette Do you see any issues with merging this now and continuing to iterate on it as needed? Is there anyone who you think needs to be alerted or given a chance to comment?

wbrunette:
I do not see an issue with merging it if it works and letting people further iterate on it. Time permitting, I will attempt to give it a try in a few days.  I cannot think of anyone who specifically needs to be tagged. I think posting to #105 and #106 have given people in the threads a chance to review and comment.

lognaturel:
Thanks, @wbrunette! We'll give you a few days to check it out if you have time and aim to merge mid next week.

@ggalmazor Could you please squash the docs commits into one when you have a chance so that we can preserve the other history? I think it's no problem to force push on a feature branch. Perhaps you could also try the App Engine variant when you have a chance? Would be good to get some of your ideas/questions on the PR filed as issues.

About BitRock InstallBuilder, how were you able to try it out, is there a trial license? @wbrunette shared the ODK OSS license with @yanokwa and we can make sure that's made available to you if you need it for verification.

ggalmazor:
Done squashing commits :)

If you want, I can create issues from my checklists after merging the PR...

Regarding BitRock, I'm using a Linux evaluation version. I don't think that I need a license to continue testing the build process yet but I'll remember to talk with @yanokwa for the official license.

ggalmazor:
I've checked that Aggregate can be built and deployed in Tomcat/PostgreSQL **and** AppEngine.

To do this I've dug deeper into the build process and I've come to the conclusion that the originally proposed subproject separation just adds more complexity without providing any real value.

After today's commits, Aggregate is a standard Gradle project.

Regarding the installer:
- It needs a WAR without config files in it because it will add into it an `ODKAggregate-settings.jar` at `WEB-INF/lib` with a new set of config files with the user's inputs from the installation wizard.
- It relied on detecting Java to run a program that would return the host's FQDN. This has proved to be buggy and it has been disabled whiel we can think of a better alternative. I have some concerns as well as to how effective we can really be detecting a host's FQDN. Now the user has to provide a valid FQDN while installing Aggregate.

wbrunette:
First and foremost, THANK YOU for taking on a monumental task of trying to upgrade and improve Aggregate! It's a large tough job.

1) While I respect you took the easiest setup first (which should be the Tomcat setup), the historic reason for Aggregate's crazy setup was that according to Mitch it needed to be that way to get the App Engine version to run. That being said, I think the last time anyone tried to adapt to Google's setup was probably 5 years ago. I started to play around to see if I could get the war to deploy in App Engine but ran out of time. It's hard to say this structure is good until it runs in App Engine Debug environment (or Google Cloud Platform) as that is where historically the inconsistencies in JavaEE come from and where most of the Aggregate debugging is needed. Did you successfully get the project running in App Engine local dev server? If so is there documentation?

2) The current setup instructions require users to have Intellij Ultimate which breaks the previous model of all ODK tools should be able to be built with free software. (Not saying that model has to stay but worth considering and is probably a larger question for the committers/TSC)

3) IMHO It would be a nicer user experience if the ant build tasks that moved the correct config files into place were replaced with custom gradle tasks. For example in the old setup the bean that controls which database was setup by a custom gradle task. Something like "gradle clean postgresConfigure build" where postgresConfigure is a custom gradle task (again my personal opinion on an improvement to simplify). I like the fact if this change did happen then we would consolidate all tasks to gradle instead of previously having some in ant and some in maven which was confusing.

4) For windows users they need to give permissions to give "full control" to users on the tomcat directory or Intellij will not have permission to expand the war in the webapps directory.
https://stackoverflow.com/questions/25579459/got-apache-tomcat-error-that-access-denied-on-this-file-localhost-access-log-201
This should probably be documented.

Things I will look into tomorrow:
1) When running the test suite with the config files set for postgres I got a failing test. Did you also get this?
2) I ran out of time to verify MySQL was working. I will test that tomorrow.

I will let @ln decide how we want to proceed. I am happy to continue further review tomorrow to try to provide more feedback. 

A possible way to proceed in a more incremental fashion is to create a branch in the opendatakit/aggregate repo that is considered in-process moving to gradle so people can have smaller pull requests that improve this wonderful work and unblock other pull requests that could be made on the new "gradle_revision" branch. Once everything is working on the "gradle_revision" branch it could then be merged to master. Not sure that is the correct approach, it's just a suggestion.

ggalmazor:
Thanks, @wbrunette, this is very useful! :)

I will look into the App Engine Debug environment. My first goal was to focus on a Tomcat+PostgreSQL configuration. I was able to run it on MySQL as well.

Regarding IntelliJ, the project is actually built with Gradle and the plan would be to have every development and build tasks supported by new Gradle tasks. There are lots of things that are missing right now like running a Gretty dev server, for example, but they'll be coming progressively.

I have that failing test under the radar (I forgot to mention it on the PR, sorry). I modified it to insert less entities into the database to run it faster but changing the `SET_SIZE` constant back to `2000` in `QueryResultTest` should do the trick.

Right now what worries me most is that it feels like I'm the only active developer working on Aggregate. If we opt to have a long-living branch and wait for feedback or help from the community I don't know if it would ever come. That and the problem of applying bug fixes to two branches instead of one.

For that reason I'd prefer to merge it on master without releasing a new version based on it. I think that this could rock the boat in a good way, getting the attention of latent contributors. It's evident that right now we would greatly benefit from people with more experience on AppEngine development while I can get to pace.

Again, thanks for your time and, for what is worth, I'm sure that more from you would be very useful. Let's keep the feedback coming :)

lognaturel:
Thank you so much for all that invaluable feedback, @wbrunette! Your experience and "historical" context is much appreciated.

I would prefer to get changes on `master` as soon as possible. This existing work helps most people who would need to build Aggregate and in particular, it's really valuable for devs on other tools to be able to quickly spin up a local Aggregate in parallel with other work. I also agree with @ggalmazor that merging things in is the most likely way to start receiving feedback. How about adding a notice at the top saying that f110aa9 was the last commit to support the previous Eclipse setup along with a link to those instructions? This might also be a good place to quickly describe the status of the migration -- postgres and mysql are confirmed working, AppEngine is not, etc.

My guess is that the only person who needs to build for AppEngine is the person preparing new releases (and testing them). Same with the BitRock installer infrastructure. Does that sound right, @wbrunette?

If you have time to try deploying a war to AppEngine that would be great @wbrunette. Alternatively/additionally, as @ggalmazor tries it, he can file issues on any problems that come up and tag @wbrunette so that he can share any hints or historical knowledge.

When it comes to IntelliJ Ultimate, I believe anyone working on the project can get a free Open Source license (just for use in an ODK context). Is that right, @yanokwa? Is the process for getting that license documented anywhere?

I propose the following course of action:
[ ] @ggalmazor adds a notice about the old Eclipse setup and the migration status at the top of the README below the badges and pushes it to this branch to update the PR
[ ] @wbrunette or @lognaturel merges this PR (with merge commit)
[ ] @ggalmazor files issues based on the conversations on this PR and shares a proposed prioritization in the dev category of the forum

wbrunette:
@ln - Unfortunately this statement is historically incorrect "My guess is that the only person who needs to build for AppEngine is the person preparing new releases (and testing them)".

Historically, Google App Engine has only supported a "subset" of java libraries. This means not all java functions or open source libraries are compatible. Here is the whitelist of classes that actually work on App Engine https://cloud.google.com/appengine/docs/standard/java/jrewhitelist It appears they support more of Java 8 than they did 7. 

The reason Aggregate's functionality has been limited was because of App Engine's limited execution environment. The good news is App Engine execution environment continues to expand it's capabilities BUT it has proven to be an false assumption that if it works on Tomcat it will work on App Engine. The general case in reverse has been more correct if it works on App Engine it will work with Tomcat, but that is also not always true either.

How Aggregate generates CSVs has always been limited by App Engine as well as how big of pictures you can upload as well as the complex logic for zipping transmissions because App Engine has not historically followed standard JavaEE. There were timeouts and other problems that have been problematic for the last 8 years. For example, there used to be a 30 second time out so all work had to be completed within that window, then they added a "background" instance for longer running tasks, then they got rid of "background" instances. It's a long story that has caused Aggregate's code base to be far more complex than it needs to be because of App Engine.

Unfortunately the majority of first time ODK users seem to really like the free App Engine model and easy install process so Aggregate has limped along.

I highly suggest that every pull request be tested on Google App Engine before acceptance otherwise the person doing the release might have to rewrite and fix a bunch of bugs as if works in Tomcat does NOT mean it will work on Google App Engine.

lognaturel:
👍 Makes sense. I believe @ggalmazor has deployed to AppEngine because he was doing some troubleshooting earlier this week but perhaps I'm remembering wrong. Let's at least get a sense of how big of a task getting it to work in that context might be before merging. I've updated the checklist above.

ggalmazor:
Alright :)

I've deployed to App Engine without any problem but I don't know if everything works as expected... I've been able to upload a form and send submissions from Collect but I don't know which features would have to be tested to really know that Aggregate is really working...

ggalmazor:
Link to the last commit with the Eclipse project added to the README file :)

wbrunette:
Since this is more of a restructure and less of a code change my guess would be basic testing in GAE would be fine. (Based on my understanding that the goal is to try to have this major restructure in master ASAP so others can work on Aggregate) 

My point really is that the future development environment (if it is going to be Intellij) should support App Engine so people can test their code changes before submitting a pull request. Also, IMHO it's easier for people to review the pull request in the development environment if there are questions about it.

Examples of basic testing:
- Upload a form
- Create a few submission
- Generate a CSV from those submissions
- Generate a KML from those submissions
- Stream the submissions to Fusion Tables
- Delete the form
- Etc

lognaturel:
Awesome! We can use that list to start a testing checklist like [the one for Collect](https://forum.opendatakit.org/t/testing-checklists/9109).

@wbrunette I'm feeling good about this going to master! Please merge when you are ready. 😊

wbrunette:
@ggalmazor were you able to generate a CSV and stream to Fusion Tables? If those work then I will be merge, those are just really common things. I am guessing they will work. I will complete the setup if necessary to check, but if you tried it already I am fine with that.

ggalmazor:
I'm trying it right now. Let me check it to be sure

lognaturel:
Thanks again for highlighting some of the issues that were likely to come up on AppEngine, @wbrunette!

@ggalmazor Last I heard you were verifying the KML export. Does that work or are there still changes needed to support that?

ggalmazor:
I’m sad ro say that exporting to KML still doesn’t work. I haven’t been able to look into that since yesterday, but today I’ll give it another go.

ggalmazor:
It turns out that with a simpler form, KML generation works perfectly. 

I've also checked that deletion of forms works too.

Looking good.

Issue #116, opened 2017-10-30
Format Audit CSV popup data into an HTML table


ggalmazor:
Newlines in the CSV content are ignored making it really hard to read, especially if there is a lot of rows in the CSV contents.

CSV data could be formatted into a nice HTML table with scrollbars for long tables.

ggalmazor:
@opendatakit-bot claim

Issue #117, opened 2017-10-30
Audit CSV popup text can't be ctrl-c copied


ggalmazor:
Right-click > copy works, though

ggalmazor:
@opendatakit-bot claim

ggalmazor:
It turns out that:
 - copying with keyboard shortcuts doesn't work but rightclick > copy works.
 - It's the modal mode what disables the keyboard shortcut to copy text

`com.google.gwt.user.client.ui.PopupPanel` says: 

```java
/**
 * Returns <code>true</code> if keyboard or mouse events that do not target
 * the PopupPanel or its children should be ignored.
 *
 * @return true if popup is modal, false if not
 */
public boolean isModal() {
  return modal;
}
```
Taking this into account, I think that two things could be happening here:
 - Either there is a bug in GWT's PopupPanel because the popup's contents are children of the PopupPanel and keyboard events should not be ignored there
 - That or we have a problem in the way those popups are built that makes GWT think that the popup's contents are not children of the popup

A quick "solution" is to disable the modal mode for the AuditCSV popup only.

ggalmazor:
@opendatakit-bot label "needs discussion"

PR #121, opened 2017-11-03
Audit csv popup improvements

ggalmazor:
This solves #116 

Snapshot:
Imgur


yanokwa:
I can confirm this PR styles the CSV data into an HTML table. It does NOT fix the issue for copying and pasting with keyboard shortcuts, but copy and paste via the contextual menu (e.g., right-click) works great. Was that the intention?

I'm on macOS 10.13.1 and I have tried Chrome 62.0.3202.89 and Safari Version 11.0.1 (13604.3.5).

ggalmazor:
Yes. I wanted to discuss in #117 the solution for the wrong copy/paste behavior.

Issue #115, opened 2017-10-30
Migrate to a Gradle task BitRock InstallBuilder project generation


ggalmazor:
The current process for generating Aggregate installers involves lots of manual work and duplicated files, making changing anything hard and prone to errors.

We could leverage a Gradle task to generate an InstallBuilder project from the source code an a set of static assets. We could also remove bulky assets like Google's AppEgine SDK that could be downloaded on-demand.

lognaturel:
A++++ did I understand correctly in Slack that you already have this partially done, @ggalmazor?

ggalmazor:
@opendatakit-bot claim

That's right. I've already prepared a branch with the solution

Issue #113, opened 2017-10-18
Installation error when running Java 1.9


yanokwa:
On macOS, installing ODK Aggregate throws a NoClassDefFoundError error.

https://forum.opendatakit.org/t/odk-aggregate-installation-error-on-mac-sierra/10232
https://forum.opendatakit.org/t/odk-aggregate-installer-failure-on-macos/9978

ggalmazor:
@opendatakit-bot claim

ggalmazor:
@yanokwa could you check if this is still happening with the new installer?

yanokwa:
I am still getting this problem. To reproduce, I did the following..

1. Put `org.gradle.java.home=/Library/Java/JavaVirtualMachines/jdk1.8.0_152.jdk/Contents/Home` in my properties so gradle uses Java 8 to compile. It can't compile with Java 9.
1. Run `./gradlew clean build -x test -PwarMode=installer && ./gradlew buildInstaller`
1. Use InstallBuilder to generate an installer.
1. Run the installer and have it launch the uploader.

<img width="504" alt="screen shot 2017-12-08 at 15 08 05" src="https://user-images.githubusercontent.com/32369/33788870-9c6e45fc-dc29-11e7-9858-2bca7f0da63f.png">

```
$ java -jar ODKAggregateAppEngineUpdater.jar
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/commons/cli/ParseException
	at java.base/java.lang.Class.forName0(Native Method)
	at java.base/java.lang.Class.forName(Class.java:375)
	at org.eclipse.jdt.internal.jarinjarloader.JarRsrcLoader.main(JarRsrcLoader.java:56)
Caused by: java.lang.ClassNotFoundException: org.apache.commons.cli.ParseException
	at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:466)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:563)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:496)
	... 3 more
```
```
$ java --version
java 9.0.1
Java(TM) SE Runtime Environment (build 9.0.1+11)
Java HotSpot(TM) 64-Bit Server VM (build 9.0.1+11, mixed mode)
Veritas:~/Desktop/ODK Aggregate >
```

yanokwa:
[...]

PR #122, opened 2017-11-03
Installer as gradle task

ggalmazor:
This solves #115 and #113


docs/installer_generation.md
```diff
@@ -2,26 +2,18 @@
...
-**Notice:** This will be replaced by gradle tasks on the main Aggregate project
-
 ## Instructions
 
-1. Delete any WAR file present at `bitrock-installer` directory
-2. Generate a WAR file without any configuration files in it using `gradle clean build -x test -Pwar_mode=installer` command
-3. copy the generated WAR file at `build/lib/aggregate-%VERSION%.war` to `bitrock-installer` directory
-4. Run `mvn clean package -f bitrock-installer/pom.xml`
-5. Open BitRock InstallBuilder app and open `bitrock-installer/buildWar.xml`
+1. Build the project for the installer build process with `gradle clean build -x test -Pwar_mode=installer` command
```
yanokwa: These should probably be `./gradlew`

docs/installer_generation.md
```diff
@@ -2,26 +2,18 @@
...
-**Maven**
-
-At the moment, the project leverages Maven to prepare it for the InstallBuilder application.
-
-**Notice:** This will be replaced by gradle tasks on the main Aggregate project
-
 ## Instructions
 
-1. Delete any WAR file present at `bitrock-installer` directory
-2. Generate a WAR file without any configuration files in it using `gradle clean build -x test -Pwar_mode=installer` command
```
yanokwa: You seem to prefer `_` to `-`. I really like hyphens because they are easier to type. Your decision whether you keep using `_`.

yanokwa:
One way forward is that we can merge just the fix for #115, roll back the changes for #113 and then try to tackle that more systematically.

docs/installer_generation.md
ggalmazor: It's because while typing, hyphens divide words as opposed to '_' that don't.

While navigating code with the keyboard `ctrl+right` and `ctrl+left` (or cmd/option alternatives in mac) per-word navigation is easier with '_'

ggalmazor:
I admit that I was a little reckless when I decided to replace that JAR that only works with Java9. Let's do what you've suggested. I'll update the PR with the old JAR.

ggalmazor:
Ok, I've reverted the update JAR that goes in the installer to the previous version.

Issue #114, opened 2017-10-30
Infinite refresh caused by data store corruption


lognaturel:
Aggregate can occasionally get into a state where it refreshes continuously. This is a known problem and a manual fix is documented at https://github.com/opendatakit/opendatakit/wiki/Aggregate-AppEngine-Troubleshooting#repairing-your-app-engine-database-yourself

This is relatively frequently reported: https://forum.opendatakit.org/t/infinite-refresh-bug/10284, https://forum.opendatakit.org/t/upload-of-aggregate-server-page-jumps/9438, https://forum.opendatakit.org/t/odk-aggregate-keeps-refreshing/6127 for example. Presumably it often goes unreported.

For blank form duplication, the problem happens immediately and the server keeps refreshing.

For submission duplication, the problem shows up when exporting to CSV and Aggregate tries to pull together that corrupt submission.

For both errors, users might sometimes see a the bad SQL query in bright red show up at the top of their screen.

At https://forum.opendatakit.org/t/odk-aggregate-wont-stop-refreshing/1264/25 Mitch says 

> The unfortunately-common-enough refresh problem is specific to Google
AppEngine and its 60-second request timeout, which aggressively terminates
the server thread, potentially in the middle of a sequence of database
writes (and often before the final set of deletes that transitions the
server to the newest version of a form definition).

https://github.com/opendatakit/aggregate/issues/79 proposes automating the db cleanup.

@yanokwa says:
> Aggregate v1.4.10 and v1.4.11 both have patches that tried to wrap the problems with mutexes.

ggalmazor:
@opendatakit-bot claim

lognaturel:
Leave him alone, @opendatakit-bot, this is a big issue! 😅

ggalmazor:
I think we can close this now that #154 has been merged...

yanokwa:
I want to hold it open until we fix all the dupe `_form` issues because I know that causes refresh.

PR #127, opened 2017-11-07
Mitigate data corruption effects like constant UI refresh

ggalmazor:
This PR acts on the repeated refresh problem of the UI due to data corruption related issues in the database. The changes include:

- Validity checks on `Form` that detect missing pieces from related `_form_info_x_y_z` tables
- Log traces for unhandled exceptions in `FormServiceImpl`

This is related to #114


src/main/java/org/opendatakit/aggregate/constants/ErrorConsts.java
```diff
@@ -26,6 +26,7 @@
  */
 public final class ErrorConsts {
 
+  public static final String POSSIBLE_CORRUPTION = "Encountered an unknown problem while reading the database. This often is a sign of database corruption. Please get the server logs and file an issue";
```
ggalmazor: This is a catch-all error for unhandled exceptions

src/main/java/org/opendatakit/aggregate/form/Form.java
```diff
@@ -713,5 +713,8 @@ public String getUri() {
     return infoRow.getUri();
   }
 
-
+  @Override
+  public boolean isValid() {
```
ggalmazor: This definition of `isValid()` would grow with more checks as we go.

src/main/java/org/opendatakit/aggregate/form/FormDefinition.java
```diff
@@ -135,9 +136,13 @@ private static final SubmissionAssociationTable getSubmissionAssociation(String
 		    		sa = st;
 		    	}
 		    }
-		}
+			}
 	    return sa;
-	}
+		} catch (Throwable t) {
```
ggalmazor: `st.getIsPersistenceModelComplete()` in line 125 will throw an `NPE` when that field is `null`. This new `try` block ensures that this exception doesn't scale. Calling code at `FormDefinition` can handle `null` `SubmissionAssociationTable` values.

src/main/java/org/opendatakit/aggregate/constants/ErrorConsts.java
lognaturel: How about adding https://github.com/opendatakit/aggregate/issues after add an issue?

ggalmazor:
Please, don't merge this PR. I'll produce a new one without the mess in git history.

ggalmazor:
This PR continues in #154

PR #151, opened 2017-11-22
Add a Configure section in README.md

### LINKED ISSUE #150: Unknown property warMode
When trying to run the `appStartWar` task for the first time I get the following error (from intelliJ or CLI):

> Error:(152, 0) Could not get unknown property 'warMode' for task ':war' of type org.gradle.api.tasks.bundling.War.

I was able to address this by moving `gradle.properties.example` to `gradle.properties`. I'm not sure whether the solution is to add that to the instructions or ...?

ggalmazor:
Closes #150

Checked in a Markdown editor that the README is correct

N/A 

None

N/A


ggalmazor:
Closing due to problems with my fork's git history. Will create it again

PR #141, opened 2017-11-15
Solution to dupes and missing rows on repeat tables

ggalmazor:
This PR solves #134

 Warning the user about missing/dupes in repeat tables would be handled in #135 

To test this PR, you can use this form: [Untitled Form.xml.tar.gz](https://github.com/opendatakit/aggregate/files/1476411/Untitled.Form.xml.tar.gz)

Just send a submission and alter the database to produce missing rows and duped rows. This example has a dupe row for `_ORDINAL_NUMBER` 1 and has no row for `_ORDINAL_NUMBER` 2.

_URI | _CREATOR_URI_USER | _CREATION_DATE | _LAST_UPDATE_URI_USER | _LAST_UPDATE_DATE | _PARENT_AURI | _ORDINAL_NUMBER | _TOP_LEVEL_AURI | SOME_TEXT
-- | -- | -- | -- | -- | -- | -- | -- | --
uuid:46666a40-dfbf-4255-bcd4-e34b33325000 | anonymousUser | 2017-11-13 22:16:31.859000 | NULL | 2017-11-15 22:16:31.859000 | uuid:f99e6bd0-e07a-4b1b-b121-a34b8c852319 | 1 | uuid:f99e6bd0-e07a-4b1b-b121-a34b8c852319 | 1
uuid:46666a40-dfbf-4255-bcd4-e34b33325001 | anonymousUser | 2017-11-14 22:16:31.859000 | NULL | 2017-11-15 22:16:31.859000 | uuid:f99e6bd0-e07a-4b1b-b121-a34b8c852319 | 1 | uuid:f99e6bd0-e07a-4b1b-b121-a34b8c852319 | 2
uuid:46666a40-dfbf-4255-bcd4-e34b33325002 | anonymousUser | 2017-11-15 22:16:31.859000 | NULL | 2017-11-15 22:16:31.859000 | uuid:f99e6bd0-e07a-4b1b-b121-a34b8c852319 | 3 | uuid:f99e6bd0-e07a-4b1b-b121-a34b8c852319 | 3

With this data, you should be able to see and export a group with rows for the text '2' and '3':

demo_repeat_group

Issue #135, opened 2017-11-11
Users should be able to fix data corruption with Aggregate


ggalmazor:
Aggregate should have a new admin section with information about data corruption in the database and automatic fixes

lognaturel:
Replaces #79. @ggalmazor it may be worth taking a quick look at some of the warnings/concerns brought up on that issue discussion. They may not all be current so take them with a grain of salt!

ggalmazor:
Ok, I'll take a look. 
(I've added a note in the issue description to warn readers that the table with the causes will change)

ggalmazor:
I'm thinking about something like this:

![](https://i.imgur.com/KpUjZdy.png)
(don't mind the "Forms without filesed grey header above the table")

In the screenshot, you can see that the last form has 2 submission association rows, which is indicative of data corruption. Therefore, the delete form button is available for that row.

The form deletion process would ensure that no orphan rows are left in other tables like `_form_info_fileset` and other like it.

ggalmazor:
@opendatakit-bot  label "needs discussion"

yanokwa:
This is a really important problem to take on!

The general approach of checking` _form_*` for missing or duplicate rows is fantastic, but I'd be consistent there. For example, you are not checking for `_form_info_manifest_bin` and it's not clear why. Is it impossible?

I also like the approach of cleaning up orphaned rows, but to me, that's a nice to have. If it ends up being a big lift, I'd delay shipping it.

I don't know if users need to know exactly why a form is broken. They just need to know that it is broken and if they can push a button to fix it. And I'd frame it this way because it's our fault that their data is corrupt. We don't want them to do extra work.

So to that end, I'd probably only list the forms that have problems. And I'd have two headers: Form ID, Problem, and then a Fix button. When you tap the Fix button, it lists the things it'll try to do, and then you can confirm.

And for things that you can't fix (e.g., a missing repeat), you maybe list the submission that has the problem and we can decide what to do later. 

Oh, and I wouldn't call it `Data Corruption Admin` because it definitely sounds like we are giving up. Maybe we put it under `Site Admin` under `Database Repair`

ggalmazor:
Alright! Thanks for the feedback :)

Knowing that we're working in the right direction I can continue adding to this new section. I think I'll be deploying to AppEngine regularly to demo it. Also, DataStore lets us be dirtier with data and reproduce more corruption scenarios :)

I have some questions/comments:

[...]

PR #153, opened 2017-11-22
Solve (partially) dupes/missing rows/wrong sequences in repeat tables

### LINKED ISSUE #134: Dupes or missing rows in repeat tables prevents users from listing and exporting submissions
There are two things that could happen inside a repeat: we have a missing row or we have a dupe row

Aggregate relies on the `_ORDINAL_NUMBER` field which should contain a sequence of ascending ints without holes and without dupes. If this sequence is wrong for 1 submission, we fail and the user can't see or export any submission at all

The safest thing to do is to show only the subset of repeat rows that match the expected sequence (1,2,3,4...) but this would lead to a rather difficult to predict behavior:
- If there is a hole as in `[1, 2, 4, 5, 6, 7]`: the user will only see rows `[1, 2]`
- If there is a dupe in the place of the hole as in `[1, 2, 2, 4, 5, 6, 7]`: the user will see `[1, 2(the first), 4, 5, 6, 7]`

The other option would be to show everything (dupes included). We would use `_ORDINAL_NUMBER` as an ordering parameter only and dupes/missing numbers wouldn't throw an error.

In any case, we wouldn't throw. Instead, we would leave a trace in the logs.

ggalmazor:
Closes #134

This PR solves a problem I introduced in a wrong rebase in PR #141 

Warning the user about missing/dupes in repeat tables would be handled in #135 

To test this PR, you can use these forms: 

- Normal repeat [Untitled Form.xml.tar.gz](https://github.com/opendatakit/aggregate/files/1476411/Untitled.Form.xml.tar.gz)
- Nested repeat [nested-repeats.xml.tar.gz](https://github.com/opendatakit/aggregate/files/1499968/nested-repeats.xml.tar.gz)

Just send a submission and alter the database to produce missing rows and duped rows. This example has a dupe row for `_ORDINAL_NUMBER` 1 and has no row for `_ORDINAL_NUMBER` 2.

_URI | _CREATOR_URI_USER | _CREATION_DATE | _LAST_UPDATE_URI_USER | _LAST_UPDATE_DATE | _PARENT_AURI | _ORDINAL_NUMBER | _TOP_LEVEL_AURI | SOME_TEXT
-- | -- | -- | -- | -- | -- | -- | -- | --
uuid:46666a40-dfbf-4255-bcd4-e34b33325000 | anonymousUser | 2017-11-13 22:16:31.859000 | NULL | 2017-11-15 22:16:31.859000 | uuid:f99e6bd0-e07a-4b1b-b121-a34b8c852319 | 1 | uuid:f99e6bd0-e07a-4b1b-b121-a34b8c852319 | 1
uuid:46666a40-dfbf-4255-bcd4-e34b33325001 | anonymousUser | 2017-11-14 22:16:31.859000 | NULL | 2017-11-15 22:16:31.859000 | uuid:f99e6bd0-e07a-4b1b-b121-a34b8c852319 | 1 | uuid:f99e6bd0-e07a-4b1b-b121-a34b8c852319 | 2
uuid:46666a40-dfbf-4255-bcd4-e34b33325002 | anonymousUser | 2017-11-15 22:16:31.859000 | NULL | 2017-11-15 22:16:31.859000 | uuid:f99e6bd0-e07a-4b1b-b121-a34b8c852319 | 3 | uuid:f99e6bd0-e07a-4b1b-b121-a34b8c852319 | 3

With this data, you should be able to see and export a group with rows for the text '2' and '3':

demo_repeat_group

### DIFF
--- a/src/main/resources/security.properties
+++ b/src/main/resources/security.properties
@@ -28,7 +28,7 @@ security.server.channelType=REQUIRES_INSECURE_CHANNEL
 # external services.
 #
 # This is configured during install.  If blank, discovers an IP address
-security.server.hostname=192.168.1.161
+security.server.hostname=localhost
 security.server.port=8080
 security.server.securePort=8443
 
--- a/src/main/resources/security.properties
+++ b/src/main/resources/security.properties
@@ -28,7 +28,7 @@ security.server.channelType=REQUIRES_INSECURE_CHANNEL
 # external services.
 #
 # This is configured during install.  If blank, discovers an IP address
-security.server.hostname=localhost
+security.server.hostname=192.168.1.161
 security.server.port=8080
 security.server.securePort=8443
 
--- a/src/main/java/org/opendatakit/aggregate/submission/SubmissionSet.java
+++ b/src/main/java/org/opendatakit/aggregate/submission/SubmissionSet.java
@@ -17,15 +17,6 @@
 
 package org.opendatakit.aggregate.submission;
 
-import java.io.PrintWriter;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
 import org.opendatakit.aggregate.datamodel.FormDataModel;
 import org.opendatakit.aggregate.datamodel.FormDataModel.DDRelationName;
 import org.opendatakit.aggregate.datamodel.FormDataModel.ElementType;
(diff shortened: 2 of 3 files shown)


ggalmazor:
It doesn't work with nested repeats. I'll look into that tomorrow

ggalmazor:
Alright! Fixed for nested repeat tables too :)

lognaturel:
🎉 @yanokwa that was the only special case I could think of. Any from you?

src/main/resources/security.properties
```diff
@@ -28,7 +28,7 @@ security.server.channelType=REQUIRES_INSECURE_CHANNEL
 # external services.
 #
 # This is configured during install.  If blank, discovers an IP address
-security.server.hostname=localhost
+security.server.hostname=192.168.1.161
```
yanokwa: This should be localhost.

PR #154, opened 2017-11-22
Prevent refresh hell on identified cases of data corruption

### LINKED ISSUE #114: Infinite refresh caused by data store corruption

ggalmazor:
Add new controls to prevent constant refresh of the UI due to unhandled exceptions server-side
Add log traces that give more information about these unhandled exceptions
Add github issues link in POSSIBLE_CORRUPTION error text

Closes #114

Comes from #127

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/form/Form.java
+++ b/src/main/java/org/opendatakit/aggregate/form/Form.java
@@ -713,5 +713,8 @@ class Form implements IForm {
     return infoRow.getUri();
   }
 
-
+  @Override
+  public boolean isValid() {
+    return filesetRow.hasField(FormInfoFilesetTable.IS_DOWNLOAD_ALLOWED);
+  }
 }
--- a/src/main/java/org/opendatakit/common/persistence/CommonFieldsBase.java
+++ b/src/main/java/org/opendatakit/common/persistence/CommonFieldsBase.java
@@ -165,6 +165,17 @@ public abstract class CommonFieldsBase {
     return Collections.unmodifiableList(fieldList);
   }
 
+  public final boolean hasField(DataField f) {
+    if (f == null) {
+      throw new IllegalArgumentException("Field value is null!");
+    }
+    if (!fieldList.contains(f)) {
+      throw new IllegalArgumentException("Attempting to get a field " + f.getName()
+          + " not belonging to " + schemaName + "." + tableName);
+    }
+    return fieldValueMap.containsKey(f);
+  }
+
   public final String getStringField(DataField f) {
     if (f == null) {
       throw new IllegalArgumentException("Field value is null!");
--- a/src/main/java/org/opendatakit/aggregate/constants/ErrorConsts.java
+++ b/src/main/java/org/opendatakit/aggregate/constants/ErrorConsts.java
@@ -26,6 +26,7 @@ package org.opendatakit.aggregate.constants;
  */
 public final class ErrorConsts {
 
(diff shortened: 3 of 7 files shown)


src/main/java/org/opendatakit/aggregate/form/Form.java
```diff
@@ -713,5 +713,8 @@ public String getUri() {
     return infoRow.getUri();
   }
 
-
+  @Override
+  public boolean isValid() {
+    return filesetRow.hasField(FormInfoFilesetTable.IS_DOWNLOAD_ALLOWED);
```
yanokwa: Why was this chosen as the measure of validity? Is it the last thing that is set before a form becomes available? Or rather, is it possible for a form to be invalid after this stage?

src/main/java/org/opendatakit/common/persistence/CommonFieldsBase.java
```diff
@@ -165,6 +165,17 @@ public final Date getLastUpdateDate() {
     return Collections.unmodifiableList(fieldList);
   }
 
+  public final boolean hasField(DataField f) {
```
yanokwa: Why not use the existing getBooleanField method?

src/main/java/org/opendatakit/aggregate/form/Form.java
ggalmazor: To be honest, it's just because that field being null is what threw an `NPE`, producing the refresh hell.

ggalmazor: This is just a temporary fix while we work on #134

src/main/java/org/opendatakit/common/persistence/CommonFieldsBase.java
yanokwa: Never mind. I think this is fine.

ggalmazor: With `getBooleanField` I would have to compare the result with `null` to decide if a form has a non-null value in that field... The semantics of `hasField(DataField f)` seem more convenient too.

ggalmazor: Anyway, everything on this PR is subject to change after #134. 

This patches only prevent some evident data corruption situations that lead to the refresh hell

PR #157, opened 2017-11-27
Enable mouse dragging to pan the map and mouse wheel for zoom

### LINKED ISSUE #156: Unable to drag map or use mouse wheel in maps
This issue was originally created by @ghost at https://github.com/opendatakit/opendatakit/issues/1281

**Software and hardware versions**
Aggregate v1.4.13 (probably all)

**Problem description**
I cannot drag or zoom the map in visualization tool in ODK Aggregate using mouse. Only zoom in and zoom out buttons in the lower right corner are working. The map is almost useless.

**Steps to reproduce the problem**
1. On the Submissions tab choose a form, which contains column of location type and submissions.
1. Press Visualize button.
1. Choose type "Map".
1. Choose "Geopoint to Map".
1. Press "Map It" button.
1. Try to drag or move the map using mouse.

**Expected behavior**
It should be possible to drag or zoom the map using mouse.

ggalmazor:
Closes #156 

Follow the instructions in #156 and verify that works as expected

This is a simple change on the options to instantiate the map

Being this so simple, I can only think that if this wasn't already in the code, maybe there was an important enough reason to not do it in the first place...

Nope.

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/client/popups/VisualizationPopup.java
+++ b/src/main/java/org/opendatakit/aggregate/client/popups/VisualizationPopup.java
@@ -432,6 +432,8 @@ public final class VisualizationPopup extends AbstractPopupBase {
     options.setMapTypeControl(true);
     options.setNavigationControl(true);
     options.setScaleControl(true);
+    options.setScrollwheel(true);
+    options.setDraggable(true);
     final MapWidget mapWidget = new MapWidget(options);
     mapWidget.setSize("100%", "100%");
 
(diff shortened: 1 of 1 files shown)


lognaturel:
Thanks @ggalmazor, @wbrunette! It was probably just a simple oversight that no one had time to address, I can't think of anything else that would explain the omission.

wbrunette:
likely an incompatibility with a web browser or app engine that is now fixed since it's 7 years later.

PR #165, opened 2017-11-28
New Database Repair tool

### LINKED ISSUE #135: Users should be able to fix data corruption with Aggregate

ggalmazor:
Closes #135

This PR adds a new Database Repair tool that lets users identify data corruption situations and apply fixes to solve them.

Development progress has been documented in #135 
There is also a [forum post](https://forum.opendatakit.org/t/odk-aggregate-future-plans/9219/10?u=ggalmazor) announcing an upcoming beta program to test this feature.


ggalmazor:
Sure!

| cause | effect | fix |
| ---- | ---- | ---- |
| missing `_form_info_fileset` | UI refresh hell | delete form |
| dupe `_form_info_fileset` | anoying alerts | delete the dupe |
| nulls in `_form_info_fileset` | UI refresh hell | fill nulls |
| missing `_form_info_submission_association` | unkown | delete form |
| dupe `_form_info_submission_association` | anoying alerts | delete the dupe |
| nulls in `_form_info_submission_association` | unkown | fill nulls |
| no `_form_data_model` for a given `_form_info_submission_association` | unknown | delete form |

If there is some combination of dupes and nulls, you can fill nulls and delete any row you want as long as one row is left with no nulls in it.

You can reproduce these data corruption scenarios by manually creating/deleting rows in your database.

yanokwa:
> _form_info_fileset

e y

ggalmazor:
Testing instructions:

1. Upload to Aggregate the following form: [simpleform.xml.tar.gz](https://github.com/opendatakit/aggregate/files/1532663/simpleform.xml.tar.gz)
1. Insert a dupe row in `_form_info_fileset` table with this SQL script:
    ```sql
    -- Create a dupe
    INSERT INTO _form_info_fileset
      SELECT
        'uuid:dupedupe-dupe-dupe-dupe-dupedupedupe' AS "_URI",
        "_CREATOR_URI_USER",
        "_CREATION_DATE",
        "_LAST_UPDATE_URI_USER",
        "_LAST_UPDATE_DATE",
        "_PARENT_AURI",
        "_ORDINAL_NUMBER",
        "_TOP_LEVEL_AURI",
        "ROOT_ELEMENT_MODEL_VERSION",
        "IS_ENCRYPTED_FORM",
        "IS_DOWNLOAD_ALLOWED",
        "LANGUAGE_CODE",
        "FORM_NAME",
        "DESCRIPTION",
        "DESCRIPTION_URL"
      FROM _form_info_fileset
      WHERE "FORM_NAME" = 'simpleform';
    ```
1. Check that the fix from commit [aee5f9](https://github.com/opendatakit/aggregate/pull/165/commits/aee5f9ead1de9f2bdbaae9fa9dafbaafb2f9c9ba) lets you remove the duplicated row
1. Update a row setting invalid `NULL` values to `IS_ENCRYPTED_FORM` and `IS_DOWNLOAD_ALLOWED` columns in filesets table with this SQL script:
    ```sql
    -- Set null fields
    UPDATE _form_info_fileset
      SET 
        "IS_ENCRYPTED_FORM" = NULL,
        "IS_DOWNLOAD_ALLOWED" = NULL
      WHERE "FORM_NAME" = 'simpleform';
    ```
1. Check that the fix from commit [4aea19](https://github.com/opendatakit/aggregate/pull/165/commits/4aea19e681793b76ce3e8c467be61d257299aae8) lets you fill those columns with a value

src/main/java/org/opendatakit/aggregate/databaseRepair/DatabaseRepairSubTab.java
```diff
@@ -0,0 +1,105 @@
...
+
+public class DatabaseRepairSubTab extends AggregateSubTabBase {
+
+  private final FlexTable reportsTable;
+  final FlowPanel reviewPanel;
+
+  public DatabaseRepairSubTab() {
+    reportsTable = new FlexTable();
+    reportsTable.addStyleName("dataTable");
+    add(new HTML("<h3>Form corruption report</h3>"));
```
yanokwa: I'd just make this Form report. Corruption is scary.

src/main/java/org/opendatakit/aggregate/databaseRepair/DatabaseRepairSubTab.java
```diff
@@ -0,0 +1,105 @@
...
+
+  private void setReportsTableHeader() {
+    reportsTable.setText(0, 0, "Form ID");
+    reportsTable.setText(0, 1, "# Filesets");
+    reportsTable.getRowFormatter().addStyleName(0, "titleBar");
+  }
+
+  private Widget buildFilesetCell(final FilesetReport fileset) {
+    if (fileset.isOk())
+      return new HTML("ok");
```
yanokwa: Typically uppercase.

yanokwa:
Couple of changes I'd like to see.
* If there is nothing to report, maybe under the title (e.g., Form report), it should say "No problems to report". And if there is nothing to review, don't show the review panel.
* Put the Review button in another column. No header is needed, but if you need one, maybe "Action"? It's hard to read when it's right next to Dupes.
* Remove is encrypted from review panel because we don't really know the implications
* Add last update date and last update user (just include the username if possible)
* Make is download allowed readonly. If null, then the fix is to set to false.
* Instead of `Mark for deletion` maybe we do `Delete on save` to make it clear that the delete will happen immediately.

ggalmazor:
> Put the Review button in another column. No header is needed, but if you need one, maybe "Action"? It's hard to read when it's right next to Dupes.

The next fixes will require more buttons... can we think about this when other fixes are added to the tool?

yanokwa:
Sure. We can punt on the buttons for now.

yanokwa:
I'm loving these changes! I'd make the headers URI, Last update date (include timestamp since that's a differentiator), Last update user, downloadable (to match the form management text), then Delete on Save.

ggalmazor:
Just wanted to point out that after the last changes, the proper way to test the fix of `NULL` value in `IS_DOWNLOAD_ALLOWED` is:
1. Update a row setting invalid `NULL` values to `IS_ENCRYPTED_FORM` and `IS_DOWNLOAD_ALLOWED` columns in filesets table with this SQL script:
    ```sql
    -- Set null fields
    UPDATE _form_info_fileset
      SET 
        "IS_DOWNLOAD_ALLOWED" = NULL
      WHERE "FORM_NAME" = 'simpleform';
    ```
1. Check that the fix from commit [4aea19](https://github.com/opendatakit/aggregate/pull/165/commits/4aea19e681793b76ce3e8c467be61d257299aae8) fills `IS_DOWNLOAD_ALLOWED` column with a `FALSE` value

src/main/java/org/opendatakit/aggregate/databaseRepair/FilesetReviewPanel.java
```diff
@@ -0,0 +1,168 @@
...
+    return markForDeletion;
+  }
+
+  private FlexTable buildTable() {
+    FlexTable table = new FlexTable();
+    table.addStyleName("dataTable");
+    table.setText(0, 0, "URI");
+    table.setText(0, 1, "Last update date");
+    table.setText(0, 2, "Last update user");
+    table.setText(0, 3, "Is download allowed?");
```
yanokwa: Just `Downloadable`

yanokwa:
If I have 3 dupes, I can't delete 1 at a time. I can only delete 2.

The workflow for fixing nulls doesn't make sense to me. Right now, I can only hit the save button if I'm deleting on save, so I can't fix the null independent of deletion?

My expectation is that if a form has Downloadable set as null, that I can fix that separately. Not sure how that UI should look like. Maybe when you hit Review, if Downloadable is null we have a button there that fixes it?

ggalmazor:
> 3 dupes
Why would you want to leave 2 dupes instead of 1 row?

> nulls wokrflow nonsense
In that case it doesn’t make sense :) I’ll make some UI changes and, of course, you have to be able to fix that even if it’s the only row there is

yanokwa:
As far as the dupes, a user might want to choose 1, do the delete/fix, run the report, etc. But @lognaturel disagrees with me, so keep it as is.

yanokwa:
Ugh, the more I think about this, the more we might be setting ourselves up for trouble.

We can't just delete any duplicate. We have to delete the duplicate that doesn't exist in the other tables. So for example, if you upload a form, the `_form_info_fileset._URI` (which is unique) is inserted into `_form_info_xform_bin._PARENT_AURI`.

The same applies to setting downloadable to null. If downloadable is null, then something happened before or after that call to set it. We should check the pipeline to make sure the tables that are created before and after are in a sensible state. That is, there is a `_form_data_model`, there is a `_form_info_submission_association`, etc, and those URIs are all consistent.

So what would be prudent is that we start with a form upload, look at how that gets mapped to the tables and when, and use that as a sanity check. That make sense, @ggalmazor?

ggalmazor:
Alright!

I'll try to understand better the form upload process and detect this kind of relations... 

In the meantime, would it help to add a column to the fileset review panel that tells if there are `_form_info_xform_bin` rows associated to each fileset row?

yanokwa:
I'm trying to think of this as a user. What would at `_form_info_xform_bin` tell me as a user that would help me decide which form to remove. If it's 0, i guess that's problem. And if it's 1, that's normal. And if it's 2, that's probably a problem (don't know if media files count).

ggalmazor:
At this stage, I think we can focus just on what gives us more info to understand a specific situation. I think users ultimately will have a "fix everything" button but for that, we need to get some knowledge first... With the work we're doing here we're already defining some heuristics that will let use automate most of the fixes.

ggalmazor:
So, this is what I've figured out about *form-fileset-xform-manifest* domain relationships:

fileset_er_model

Apart from having valid relationships between these entities, there are other requirements that have to be met in order to consider that a form is valid:

| Domain | Rules |
| ------ | ----- |
| A form | <ul><li>It has exactly 1 valid `_form_info_fileset` row</li><li>It has exactly 1 valid row in all three `_form_info_xform_*` tables</li></li><li>When it has attachments:<ul><li>Each attachment has exactly 1 valid row in all three `_form_info_manifest_*` tables</li><li>The values in `_form_info_manifest_bin._ORDINAL_NUMBER` represent a valid sequence (no duped or missing numbers, starting with 1)</li></ul></ul> |
| A fileset | <ul><li>`_ORDINAL_NUMBER` contains a `1`</li><li>`IS_ENCRYPTED_FORM` is non-null (it contains a boolean)</li><li>`IS_DOWNLOAD_ALLOWED` is non-null (it contains a boolean)</li></ul> |
| A xform BIN | <ul><li>`_ORDINAL_NUMBER` contains a `1`</li><li>`UNROOTED_FILE_PATH` is null</li><li>`CONTENT_TYPE` contains `text/xml`</li><li>`CONTENT_LENGTH` contains a positive integer</li></ul> |
| A xform BLB | <ul><li>`VALUE` is non-empty (it contains the form in XML format)</li></ul> |
| A xform REF | <ul><li>`PART` contains a `1`</li></ul> |
| A manifest BIN | <ul><li>`UNROOTED_FILE_PATH` is non-empty (it contains a filename)</li><li>`CONTENT_TYPE` is non-empty (it contains the attachment's MIME type string)</li><li>`CONTENT_LENGTH` contains a positive integer</li></ul> |
| A manifest BLB | <ul><li>`VALUE` is non-empty (it contains the attachment)</li></ul> |

I think we could safely ignore some rules because it doesn't seem that they really affect Aggregate's behavior: 
- `ORDINAL_NUMBER` value in `_form_info_fileset` (there is always 1 row)
- all rules for `_form_info_xform_bin` (it's always an XML document) 
- `PART` in `_form_info_xform_ref` (there is always 1 row)

lognaturel:
Y'all have a plan for making sure all that great info isn't lost in a closed PR, right?!

ggalmazor:
@lognaturel yeah :) we will create a new issue with my last comment as the blueprint for the database report

PR #166, opened 2017-11-28
Support versioning like in Briefcase

ggalmazor:
**Adds artifact versioning system from Briefcase**

Now WAR files are versioned following the same scheme as in Briefcase.

**Adds a notice about the running version inside the HelpBoopPopup**

The version is written into a `BuildConfig` class by Gradle and it's printed inside the Help popup

snapshot

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -304,8 +304,9 @@ task buildInstaller(dependsOn: [cleanInstaller, gaeDownloadSDK], type: Copy) {
   }
 
   doLast {
-    file("build/installer/files/${archivesBaseName}-${version}.war").renameTo(file("build/installer/files/ODKAggregate.war"))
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -164,7 +180,7 @@ war {
 
 gretty {
   contextPath = '/'
-  host = 'localhost'
+  host = '192.168.1.161'
```
yanokwa: Should be localhost.

yanokwa:
I think the version information should be under Site Admin -> Preferences > Version Information and only there. Eventually, we'll want to put the build number on every page, but for now, Version Information is the right spot.

ggalmazor:
That makes much more sense.

PR #172, opened 2017-12-05
Issue 171 code style normalization

### LINKED ISSUE #171: Normalize code style formatting
Currently, there is a mix of styles in the code, including tabs and spaces for indenting code. This makes reviewing changes harder as GitHub detects more changed lines than the ones that actually have the important changes in the code that must be reviewed.

To fix this, we could run a massive reformat action using IntelliJ and add to the repo the actual IntelliJ settings file with the used code formatting settings.

Any

#### Problem description
Mix of code styles make change reviews harder

N/A

#### Expected behavior
N/A

#### Other information 
N/A

ggalmazor:
Closes #171

This is a safe, automatic code reformat performed by IntelliJ. Nevertheless, I've checked that the project compiles and runs.

It is what makes sense since we use IntelliJ as the official IDE for Aggregate

This should be a safe change because code formatting is done by the IDE automatically

N/A

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,6 +19,7 @@ plugins {
   id 'maven'
   id 'war'
   id 'idea'
+  id 'checkstyle'
   id 'de.fuerstenau.buildconfig' version '1.1.8'
 }
 apply plugin: 'gwt'
@@ -317,3 +318,7 @@ buildConfig {
   packageName = 'org.opendatakit.aggregate.buildconfig'
 }
 
+// Required to use fileExtensions property in checkstyle file
+checkstyle {
+  toolVersion = '7.6.1'
+}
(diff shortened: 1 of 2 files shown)


ggalmazor:
Let me check that I understood correctly: in this PR, you want just a commit with the reformatting of the code?

Then we would add the Gradle checkstyle task and a commented out checkstyle rules file in a new PR, or in this one as a second commit?

yanokwa:
Reverse it. I'd like a PR with just the checkstyle commit first with all the rules commented out. Then there can be follow on PRs that uncomment one of those rules and have the fixes for that one rule.

yanokwa:
Reverse it. I'd like a PR with just the checkstyle commit first with all the rules commented out. Then there can be follow on PRs that uncomment one of those rules and have the fixes for that one rule.

PR #175, opened 2017-12-08
Add CI support

yanokwa:
Confirmed that the build produces reports and libs.

We use CircleCI and we need it on all our tools. This PR does not test the code, but that will come eventually. 

None.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ jobs:
       - run:
           name: Configure build
           command: cp gradle.properties.example gradle.properties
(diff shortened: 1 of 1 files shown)


ggalmazor:
Does this change build just master or any branch that gets changed? How hard would it be to build on CircleCI any branch that has an active PR on it?

yanokwa:
By default, CircleCI builds all commits. I believe this is regardless of branch. 

I've also enabled building forked pull requests (https://circleci.com/gh/opendatakit/aggregate/edit#advanced-settings) so we can run tests on PRs.

PR #181, opened 2017-12-16
Add Docker support for a development PostgreSQL server

ggalmazor:
Instructions to run the dev server are in the [new section on the README.md](https://github.com/ggalmazor/aggregate/tree/docker_dev_postgresql#postgresql-with-docker-using-docker-compose).

This change will allow us to add other platforms like MySQL as well.

PR #189, opened 2018-01-12
Enable testing on CircleCI

### LINKED ISSUE #188: Set Up Test Summary on CircleCI builds
CircleCI runs the tests and fails if any test fails but we don't get a summary of what's going on with each build attempt.

N/A

#### Problem description
N/A

N/A

#### Expected behavior
N/A

#### Other information 
N/A

yanokwa:
Closes #188 

I've run this on my branch and the tests run, but one test (org.opendatakit.common.persistence.QueryResultTest) fails, so the build fails. 

I've used default variables (e.g., circleci) that are the default so there is less "code" in the config. I used full switches for readability. The file location for the init.sql was done to match #181. Maybe it should be at root level?

Yes, this breaks the build. And because the build is broken, I'm cannot guarantee `.gradlew war` will actually produce a war. I bet it will though. Another risk is that I've grouped the gradle commands and we may run into memory problems. A final risk is that the test output is enormous and it makes loading the log in CircleCI annoying. Maybe we can silence that?

#### Remaining
- [x] Fix for QueryResultTest. Can you help here, @ggalmazor 
- [x] Confirm we get a war
- [x] Confirm we get a test artifacts/reports
- [x] Agree on a location for init.sql and update .gitignore

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,13 +4,17 @@ jobs:
     working_directory: ~/src
     docker:
       - image: circleci/openjdk:latest
+      - image: circleci/postgres:9.6
+        environment:
+        - POSTGRES_USER=circleci
+        - POSTGRES_DB=circleci
     steps:
       - checkout
       - run:
-          name: Install checkout requirements
+          name: Install requirements
           command: |
             curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | sudo bash
-            sudo apt-get install git-lfs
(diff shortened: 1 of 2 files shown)


ggalmazor:
Ok, I'll take a look at `QueryResultTest` :)

ggalmazor:
@opendatakit-bot claim

yanokwa:
I'm thinking that we should keep the init.sql separate because we might want different credentials. If that's OK with you, then we can merge this.

ggalmazor:
Alright :)

PR #193, opened 2018-01-22
Fix QueryResultTest

ggalmazor:
Related to PR #189

This PR changes one assertion to use safer values, now that the test inserts much fewer fixtures than before.

Run the test. See it pass.

Honestly, I don't quite understand what this test is trying to achieve. The failing assertions were testing rows paging capabilities that could be tested on a much simpler test.

Nope

Nope.


src/test/java/org/opendatakit/common/persistence/QueryResultTest.java
```diff
@@ -361,10 +363,21 @@ public void testCase3() throws ODKDatastoreException {
         int fetchSizes[] = {1, 2, 1021, 303, 101, 2831 };
```
yanokwa: Agreed that this test is kinda weird. Since `fetchSizes` and `idxFetch` are used elsewhere, I would have expected a more narrow fix to be where you add 10 to the `fetchSizes[]` and update `idxFetch` to 5. Then maybe add a comment that about this being governed by SET_SIZE up above.

yanokwa: Agreed that this test is kinda weird. Since `fetchSizes` and `idxFetch` are used elsewhere, I would have expected a more narrow fix to be where you add 10 to the `fetchSizes[]` and update `idxFetch` to 5. Then maybe add a comment that about this being governed by SET_SIZE up above.

src/test/java/org/opendatakit/common/persistence/QueryResultTest.java
```diff
@@ -361,10 +363,21 @@ public void testCase3() throws ODKDatastoreException {
         int fetchSizes[] = {1, 2, 1021, 303, 101, 2831 };
         int idxFetch = 4;
         int len = 0;
-        QueryResult result = query.executeQuery(null, fetchSizes[idxFetch]);
+        // It seems that the following two asserts try to test if the results
```
yanokwa: I think the comment could have been worded better. What about this?

```
// It seems that the following two asserts try to test if the results
// have less rows than the ones retrieved from the database by the query
// This was the original query:
// QueryResult result = query.executeQuery(null, fetchSizes[idxFetch]);
// This is the modified query ensuring that the page size is less than
// the number of rows in the database
```

src/test/java/org/opendatakit/common/persistence/QueryResultTest.java
```diff
@@ -361,10 +363,21 @@ public void testCase3() throws ODKDatastoreException {
...
+        // This was the original query:
+        // QueryResult result = query.executeQuery(null, fetchSizes[idxFetch]);
+        // This is the modified query ensuring that the page size is inferior to
+        // the number of rows on the database
+        QueryResult result = query.executeQuery(null, 10);
         len += result.getResultList().size();
         assertEquals( true, result.hasMoreResults());
-        assertEquals( fetchSizes[idxFetch], result.getResultList().size());
+        // This was the original assertion, which dependened on some fixed
+        // values having sense in relation to what's inserted as test fixtures
```
yanokwa: I don't understand this part. `values having sense in relation to what's inserted as test fixtures`

src/test/java/org/opendatakit/common/persistence/QueryResultTest.java
```diff
@@ -361,10 +363,21 @@ public void testCase3() throws ODKDatastoreException {
...
+        // by the query
+        // This was the original query:
+        // QueryResult result = query.executeQuery(null, fetchSizes[idxFetch]);
+        // This is the modified query ensuring that the page size is inferior to
+        // the number of rows on the database
+        QueryResult result = query.executeQuery(null, 10);
         len += result.getResultList().size();
         assertEquals( true, result.hasMoreResults());
-        assertEquals( fetchSizes[idxFetch], result.getResultList().size());
+        // This was the original assertion, which dependened on some fixed
```
yanokwa: Should be `which depends on some fixed`

src/test/java/org/opendatakit/common/persistence/QueryResultTest.java
ggalmazor: Sorry about that... `english.exe` had stopped working last night.

lognaturel:
@ggalmazor Do you intend to keep the old `QueryResultTest` or just replace it with your new one? If I'm not mistaken the new one provides approximately the same coverage but much more clearly.

ggalmazor:
@lognaturel yes, I want to eventually remove it but first I wanted to give you some time to check that I didn't miss anything important

src/test/java/org/opendatakit/common/persistence/NewQueryResultTest.java
```diff
@@ -0,0 +1,231 @@
...
+    element.setLongField(TestTable.integerField, row.integerField.longValue());
+    element.setNumericField(TestTable.doubleField, row.doubleField);
+    element.setDateField(TestTable.dateField, row.dateField);
+    element.setBooleanField(TestTable.booleanField, row.booleanField);
+
+    ds.putEntity(element, user);
+  }
+
+  private static void insertTestFixtures() throws ODKEntityPersistException, ODKOverQuotaException {
+    System.out.println("Inserting " + TEST_ROWS.length + " test rows");
```
nribeka: Is the sysout here intentional?

src/test/java/org/opendatakit/common/persistence/QueryResultTest.java
```diff
@@ -14,6 +14,9 @@
 package org.opendatakit.common.persistence;
 
 import static org.junit.Assert.assertEquals;
+import static org.opendatakit.common.persistence.Query.Direction.*;
```
nribeka: I am not a big fan of star import :)

ggalmazor: I'm curious about this... why? IntelliJ by default does this for you whenever you import more than 3 or 5 members from the same class...

dcbriccetti: I like star import in certain cases, such as with an enum when you’re using all the values.

src/test/java/org/opendatakit/common/persistence/QueryResultTest.java
```diff
@@ -361,10 +363,20 @@ public void testCase3() throws ODKDatastoreException {
         int fetchSizes[] = {1, 2, 1021, 303, 101, 2831 };
         int idxFetch = 4;
         int len = 0;
-        QueryResult result = query.executeQuery(null, fetchSizes[idxFetch]);
+        // It seems that the following two asserts try to test if the results
+        // have less rows than the ones retrieved from the database by the query
```
dcbriccetti: A period at the end of this sentence would be nice.
(I no longer worry about the less/fewer distinction.)

src/test/java/org/opendatakit/common/persistence/QueryResultTest.java
```diff
@@ -361,10 +363,20 @@ public void testCase3() throws ODKDatastoreException {
...
+        // This was the original query:
+        // QueryResult result = query.executeQuery(null, fetchSizes[idxFetch]);
+        // This is the modified query ensuring that the page size is less than
+        // the number of rows in the database
+        QueryResult result = query.executeQuery(null, 10);
         len += result.getResultList().size();
         assertEquals( true, result.hasMoreResults());
-        assertEquals( fetchSizes[idxFetch], result.getResultList().size());
+        // This was the original assertion, which depends on fetching less
+        // rows than those that have been inserted at the @BeforeClass setup method
```
dcbriccetti: A colon at the end would be very nice.

src/test/java/org/opendatakit/common/persistence/TestRow.java
```diff
@@ -0,0 +1,26 @@
...
+
+class TestRow {
+    public final String stringField;
+    public final Integer integerField;
+    public final WrappedBigDecimal doubleField;
+    public final Date dateField;
+    public final Boolean booleanField;
+
+    TestRow(String someString, Integer someInteger, Double someDouble, String someDate, Boolean someBoolean) {
+        this.stringField = someString;
```
dcbriccetti: If you use the `some`, how about removing the `this.`s.

src/test/java/org/opendatakit/common/persistence/TestRow.java
```diff
@@ -0,0 +1,26 @@
...
+        this.integerField = someInteger;
+        this.doubleField = (someDouble == null) ? null : WrappedBigDecimal.fromDouble(someDouble);
```
dcbriccetti: Any thought to Optional here, if this field is optional?

src/test/java/org/opendatakit/common/persistence/NewQueryResultTest.java
```diff
@@ -0,0 +1,231 @@
+/**
```
dcbriccetti: Is this new code? Is this the right notice?

src/test/java/org/opendatakit/common/persistence/NewQueryResultTest.java
```diff
@@ -0,0 +1,231 @@
...
+import org.opendatakit.common.persistence.exception.ODKDatastoreException;
+import org.opendatakit.common.persistence.exception.ODKEntityPersistException;
+import org.opendatakit.common.persistence.exception.ODKOverQuotaException;
+import org.opendatakit.common.security.User;
+import org.opendatakit.common.web.CallingContext;
+import org.opendatakit.common.web.TestContextFactory;
+
+
+public class NewQueryResultTest {
+  private static final int SET_SIZE = 3;
```
dcbriccetti: Move this down, and set the value from the length of one of the arrays?

src/test/java/org/opendatakit/common/persistence/NewQueryResultTest.java
```diff
@@ -0,0 +1,231 @@
...
+  private static TestRow[] TEST_ROWS = new TestRow[STRINGS.length * SET_SIZE * DOUBLES.length * DATES.length * BOOLEANS.length];
+
+  static {
+    int i = 0;
+    for (String someString : STRINGS)
+      for (int someInteger = 0; someInteger < SET_SIZE; someInteger++)
+        for (Double someDouble : DOUBLES)
+          for (String someDate : DATES)
+            for (Boolean someBoolean : BOOLEANS)
+              TEST_ROWS[i++] = new TestRow(someString, someInteger, someDouble, someDate, someBoolean);
```
dcbriccetti: This is a cool approach to generating data.

src/test/java/org/opendatakit/common/persistence/TestTable.java
```diff
@@ -0,0 +1,55 @@
+package org.opendatakit.common.persistence;
+
+import static org.opendatakit.common.persistence.DataField.DataType.BOOLEAN;
+import static org.opendatakit.common.persistence.DataField.DataType.DATETIME;
+import static org.opendatakit.common.persistence.DataField.DataType.DECIMAL;
+import static org.opendatakit.common.persistence.DataField.DataType.INTEGER;
+import static org.opendatakit.common.persistence.DataField.DataType.STRING;
+
+import org.opendatakit.common.persistence.exception.ODKDatastoreException;
```
dcbriccetti: Check for unused imports

src/test/java/org/opendatakit/common/persistence/NewQueryResultTest.java
```diff
@@ -0,0 +1,231 @@
...
+      insertRowSets(table, row);
+      if (++n % 100 == 0)
+        System.out.println("Inserted " + n + " rows");
+    }
+    System.out.println("Done writing the large dataset");
+  }
+
+  private static void dropTable() throws ODKDatastoreException {
+    System.out.println("Dropping the test table");
+    TestTable table = createTestTable();
```
dcbriccetti: Odd that we have to create a table to drop the table, but I see that `dropRelation` requires (but doesn’t use) more than just the schema name and table name.

yanokwa:
Thanks, Dave! @ggalmazor can you send in a PR to remove the old test and then we can merge!

src/test/java/org/opendatakit/common/persistence/NewQueryResultTest.java
ggalmazor: These souts where previously being used and I'm up for removing them.

I think they made sense to have a little feedback while we were inserting a bazillion rows and the test was running for several minutes. Now we insert 200ish rows and it runs on seconds...

src/test/java/org/opendatakit/common/persistence/QueryResultTest.java
ggalmazor: Honestly I didn't think about this before. I don't really care because IntelliJ will handle it. I suggest we let checkstyle make our lives miserable on a future PR :P

src/test/java/org/opendatakit/common/persistence/NewQueryResultTest.java
ggalmazor: That's a cartesian product with Vavr.io...

src/test/java/org/opendatakit/common/persistence/NewQueryResultTest.java
ggalmazor: Down where? do you mean the order of members is not right? Do we have a convention for that?

Regarding deriving the `SET_SIZE` value from the other array's length, I don't want to couple those things... It's just a coincidence that `SET_SIZE` is `3` and the arrays have 3 items...

src/test/java/org/opendatakit/common/persistence/NewQueryResultTest.java
ggalmazor: Everything related to GAE and DataStore is odd, if you ask me...

PR #202, opened 2018-02-23
Packer integration

ggalmazor:
This PR brings Packer support through some new Gradle tasks.

The new Gradle task `packerBuild` will produce an OVA file with:
- Tomcat 8.0
- PostgreSQL 9.6
- ODK Aggregate deployed as ROOT webapp

The VM will ask the user to reset `root` user's password upon the first login.

Recommended usage: `./gradlew clean build packerBuild -xtest`

## VM - Published ports
The VM will serve via NAT the following services:

| Service | Host port | Guest port |
| --- | --- | --- |
| HTTP | 10080 | 80 |
| HTTPS | 10443 | 443 |

## New build parameter: `packer`

The build process needs to download the Packer binary. The URL must be configured on `gradle.properties` like so:

```properties
packer=https://releases.hashicorp.com/packer/1.2.1/packer_1.2.1_linux_amd64.zip
```

It's important that the URL corresponds to a version greater or equal than 1.2.1 and the correct OS (linux, mac or windows).


packer/aggregate.json
```diff
@@ -0,0 +1,218 @@
...
+      "execute_command": "echo 'vagrant' | {{.Vars}} sudo -S -E sh -eux '{{.Path}}'",
+      "expect_disconnect": true,
+      "scripts": [
+        "scripts/update.sh",
+        "scripts/sshd.sh",
+        "scripts/networking.sh",
+        "scripts/sudoers.sh",
+        "scripts/vagrant.sh",
+        "scripts/systemd.sh",
+        "scripts/aggregate/ansible.sh"
```
yanokwa: Since we are going our own way, we should move this to scripts/ansible.sh

packer/aggregate.json
```diff
@@ -0,0 +1,218 @@
...
+        "http_proxy={{user `http_proxy`}}",
+        "https_proxy={{user `https_proxy`}}",
+        "no_proxy={{user `no_proxy`}}"
+      ],
+      "execute_command": "echo 'vagrant' | {{.Vars}} sudo -S -E sh -eux '{{.Path}}'",
+      "expect_disconnect": true,
+      "scripts": [
+        "scripts/virtualbox.sh",
+        "scripts/vmware.sh",
+        "scripts/aggregate/cleanup.sh",
```
yanokwa: Move this to scripts/cleanup.sh

packer/aggregate.json
```diff
@@ -0,0 +1,218 @@
...
+    "iso_checksum": "83480be837710a76fd4e75a6573ca110e06f5a7589d2d3852bdb0f45749800b3",
+    "iso_checksum_type": "sha256",
+    "iso_name": "debian-9.3.0-amd64-netinst.iso",
+    "memory": "1024",
+    "mirror": "http://cdimage.debian.org/cdimage/release",
+    "mirror_directory": "9.3.0/amd64/iso-cd",
+    "name": "aggregate",
+    "networking": "bridged",
+    "no_proxy": "{{env `no_proxy`}}",
+    "preseed_path": "aggregate/preseed.cfg",
```
yanokwa: Move this to http/preseed.cfg

packer/scripts/metadata.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh -eux
```
yanokwa: We don't need this file.

packer/scripts/parallels.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh -eux
```
yanokwa: We don't need this file

packer/ansible/roles/aggregate/files/motd/10-services
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
```
yanokwa: Given that Tomcat and PostgreSQL are not configurable, does the user really need to know that both are running on so and so port?

ggalmazor: I'll remove it on my next changeset with gradle tasks :) It's almost ready

ggalmazor:
Take a look @yanokwa 

Configure correctly the value on `packerZip` in `gradle.properties` and then execute `./gradlew clean build packerBuild -x test -PwarMode=complete`

packer/ansible/roles/aggregate/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+    install_recommends: no
+  with_items:
+    - unzip
+
+- name: set facts
+  set_fact:
+    tomcat_webapps_dir: /var/lib/tomcat8/webapps
+    motd_dir: /etc/update-motd.d
+
+- name: setup database - create user odk
```
yanokwa: I'd prefer we use aggregate as the username and the database. This makes it consistent with the older VMs and I think it's clearer than `odk`.

My gut also says we should source a single sql file to create the DB. If we can get it from the source, it'd be good so we don't have to change this later.

packer.gradle
```diff
@@ -0,0 +1,61 @@
...
+
+task packerPrepareProject(dependsOn: [packerClean, packerDownloadPacker], type: Copy) {
+  into "${buildDir}/packer"
+  from fileTree(dir: 'packer')
+
+  into('ansible/roles/aggregate/files/ROOT') {
+    from zipTree("${buildDir}/libs/${archivesBaseName}-${version}.war")
+  }
+
+  doLast {
```
yanokwa: I like `aggregate` as username password and schema because `odk` is a bit vague.

packer/ansible/roles/aggregate/tasks/main.yml
ggalmazor: I'm no Ansible expert and I didn't find a way to put an inlined `<<< EOF` like we do with the dev docker database. Then I saw some guides that recommended to place each SQL sentence separately.

How would you do this? Transfer a file to the VM, run it and then remove it? Any other way?

yanokwa: Yeah, I usually put it in files/, then copy it over to /tmp on the VM, then run it. No need to remove because it will go away during packaging.

If you have links to those guides, please share. What you found might be the best practice!

packer/http/preseed.cfg
```diff
@@ -0,0 +1,48 @@
...
+d-i partman/confirm_write_new_label boolean true
+d-i passwd/root-login boolean false
+d-i passwd/root-password-again password vagrant
+d-i passwd/root-password password vagrant
+d-i passwd/user-fullname string vagrant
+d-i passwd/user-uid string 1000
+d-i passwd/user-password password vagrant
+d-i passwd/user-password-again password vagrant
+d-i passwd/username string vagrant
+d-i pkgsel/include string sudo wget curl make openssh-server
```
yanokwa: My vision for preseed.cfg was that we only install the barest minimum tools we needed to get the thing to build. There are other tools at https://github.com/chef/bento/blob/master/debian/http/debian-8/preseed.cfg#L33 that I removed for that reason.

Upon review, I think we should add `acpid` after sudo because I think it's needed to handle power on/off signals in the hypervisor. See https://unix.stackexchange.com/questions/73877/do-i-need-acpid-in-a-virtual-machine

ggalmazor:
Work is coming along, @opendatakit-bot, don't be impatient! :P

ggalmazor:
@yanokwa so, according to some users on SO, we can add this to the `vboxmanage_post` section of the packer build file to achieve port forwarding:

```json
[
  "modifyvm",
  "{{.Name}}",
  "--natpf1",
  "guest_ssh,tcp,,10022,,22"
],
[
  "modifyvm",
  "{{.Name}}",
  "--natpf1",
  "guest_ssh,tcp,,10080,,80"
],
[
  "modifyvm",
  "{{.Name}}",
  "--natpf1",
  "guest_ssh,tcp,,15432,,5432"
]
```

This would open ports 22, 80 and 5432 and map them to their corresponding 1xxxx port on the host.

Would this be enough?

ggalmazor:
I wanted to confirm what we're trying to achieve with the networking configuration script we want to create, 

This script should be used after importing the Aggregate VM and it would help our users to set up an FQDN for their Aggregate instance. It would also help configuring the ports used by Aggregate in case users make any customization to the default VM settings.

I understand that the script for this would need to do these tasks:
- Stop Tomcat
- Edit `security.properties` and set:
  - `security.server.hostname` to whatever the user wants to use:
    - A public IP address, bound to the host machine
    - A domain name bound somehow to the host machine's IP address
  - `security.server.port` and `security.server.securePort` to whatever ports are being used by Aggregate
- Start Tomcat

@yanokwa how does this sound to you? Am I missing something?

yanokwa:
Re: port forwarding. I think that port forwarding should be enough, but we should try it and see if it actually works on the packer build. I'd only forward 80 and 443 to start. I think exposing 5432 would be bad since there are weak passwords there. And I shy away from exposing 22 unless you really think it's needed.

Re: networking configuration. Yes, your understanding of the script is clear. If they change their ports, they'll have to change their port mapping, so we should also print a message telling them that.

ggalmazor:
Just a progress report of the config script I'm working on, @yanokwa.

The script will be shipped with the VM and located at `/usr/local/bin/aggregate-config` (on $PATH). The help message says this:

```
Usage: aggregate-config <flags> [<arguments>...]

Flags:
--help                  Show this help message

Arguments:
--fqn        <value>    Set a new FQN (fully qualified name)
--fqn        auto       Enable automatic detection of FQN
                        (not recommended for production environments)
--http-port  <value>    Set a new HTTP port
--https-port <value>    Set a new HTTPS port
```

- Executing it without args will throw an error asking the user to define at least one argument and showing the help. It will exit with a non-zero status
- The FQN can be set to any value or to a special `auto` value that will leave `security.server.hostname` empty to let Aggregate figure that one out
- The config tool will show you the final settings and it will stop and start Tomcat for you

The final user will have to log into the VM using ssh and execute this command to update Aggregate conf whenever she changes NAT settings on VirtualBox.

packer/ansible/roles/aggregate/files/aggregate-config.sh
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env bash
```
yanokwa: s/fqn/fqdn/g

packer/ansible/roles/aggregate/files/aggregate-config.sh
```diff
@@ -0,0 +1,101 @@
...
+set -- "${POSITIONAL[@]}"
+
+help() {
+  echo "Usage: aggregate-config <flags> [<arguments>...]"
+  echo ""
+  echo "Flags:"
+  echo "--help                  Show this help message"
+  echo ""
+  echo "Arguments:"
+  echo "--fqn        <value>    Set a new FQN (fully qualified name)"
```
yanokwa: fully qualified domain name

packer/ansible/roles/aggregate/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
+---
```
yanokwa: You only need quotes when there is a {{ variable }}. And for consistency, let's put single quotes around those variables.

packer/ansible/roles/aggregate/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
...
+    dest: '{{ motd_dir }}'
+    mode: 0755
+
+- name: install config tool
+  copy:
+    src: 'aggregate-config.sh'
+    dest: '/usr/local/bin/aggregate-config'
+    mode: 0755
+
+- name: enable ssh access for root
```
yanokwa: Is there something magical happening here? It looks like it's just echoing text.

packer/ansible/roles/aggregate/files/motd/10-services
```diff
@@ -0,0 +1,22 @@
...
+
+PROPS_FILE=/var/lib/tomcat8/webapps/ROOT/WEB-INF/classes/security.properties
+
+rawHostname=$(grep hostname ${PROPS_FILE} | awk '{split($0,a,"="); print a[2]}')
+hostname=${rawHostname:-localhost}
```
yanokwa: It would be really nice to print the IP of enp0S3 or whatever the non-loop back IP address was. That's the IP that users who aren't NATed would use. Users shouldn't have to run `ip address` to figure that out.

yanokwa: Also, I think we should move this to /etc/issue so that this information prints without users having to login.

packer/ansible/roles/aggregate/files/motd/00-header
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
yanokwa: I think maybe we should drop this. I know it's fun, but it doesn't add much.

yanokwa:
You know, I'm liking this defaulting to NAT thing, because it's by default safe. 

Then that means if you want to connect Collect, you need to switch to bridged, login as root (which forces you to change the password), then run `aggregate-config --fqdn auto` and then you are done.

packer/ansible/roles/aggregate/files/aggregate-config.sh
```diff
@@ -0,0 +1,101 @@
...
+  echo "--fqn        auto       Enable automatic detection of FQN"
+  echo "                        (not recommended for production environments)"
```
yanokwa: I don't think that's true. If you have DHCP, it's perfectly fine to use auto for production. Maybe instead, `Not recommended for dynamic IP addresses`

packer/ansible/roles/aggregate/files/aggregate-config.sh
```diff
@@ -0,0 +1,101 @@
...
+echo "New settings are:"
+grep hostname ${PROPS_FILE}
+grep port ${PROPS_FILE}
+grep securePort ${PROPS_FILE}
+echo ""
+
+echo "Starting Tomcat. Please wait..."
+echo ""
+service tomcat8 start
+echo "Done"
```
yanokwa: It'd nice to print the /etc/issue after this change. That is, when I set auto, the next thing I need to know is what IP I should use to make sure it's working.

packer/ansible/roles/aggregate/tasks/main.yml
ggalmazor: You're right, fixing it!

packer/ansible/roles/aggregate/files/motd/10-services
ggalmazor: This is the second version of this script. I'll rescue some code from the previous version which does just that.

We show this message now:
```
Open a browser and use one of these URLs to access ODK Aggregate
http://cocotero.com:10080
https://cocotero.com:10443

This VM uses these IP addresses:
10.0.2.15
```

The nice thing about MOTD is that it has a specific system to support dynamic MOTDs.

`issue`, while it lets you insert some basic stuff, it's not so nice to make it dynamic like we want. We can make a script that writes the contents of `/etc/issue`, but I don't see an easy way to detect networking changes and hook the script to them to update the `issue` file whenever there's a change.

I've made some tests and tried placing the script on `/etc/network/if-up.d` and `/etc/network/if-post-down.d` but it doesn't detect when I change the VM's network interface on a live machine.

What I've done is to add the script to `/etc/rc.local` to have it change the `issue` message while booting.

Then `aggregate-config` runs the update script at the end. It's not perfect but I think that it could be enough... what do you think?

ggalmazor:
>Then that means if you want to connect Collect, you need to switch to bridged, login as root (which forces you to change the password), then run aggregate-config --fqdn auto, reboot, then you are done.

I think you could get away with that without switching to bridged mode by setting the FQDN to whatever is valid *for the host machine*.

>Then we don't need any extra scripts or remote SSH.

I've removed the fix for remote ssh root access, but bear in mind that users with non-English keyboard layouts could have trouble using VirtualBox's console. Using SSH is easier in such scenario because it lets you use your layout, not the machine's default one.

ggalmazor:
I've rebased the branch with master and applied all the feedback :)

I've manually tested the OVA on my machine and it's looking good so far :)

PR #205, opened 2018-03-08
Improve logging

ggalmazor:
This PR adds some dependencies to the project to be able to **migrate all logging code to SLF4J**

It includes the JUL binding as a default and comes also with JCL and Log4J to SLF4J bridges, meaning that:
- We write code that uses SLF4J API directly
- Third party libraries that use Log4J or JCL get routed to SLF4J
- JUL is what ultimately writes log lines however it's configured on the `logging.properties` files.

This PR also reviews the logging code around the `UploadSubmissionsWorkerImpl` class and the Google API external service classes.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -99,7 +99,16 @@ def versions = [
     mockito       : '2.+'
 ]
 
+configurations.all {
+  exclude group: 'commons-logging', module: 'commons-logging'
+}
+
 dependencies {
+  compile group: 'org.slf4j', name: 'slf4j-api', version: '1.7.25'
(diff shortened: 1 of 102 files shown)


yanokwa:
Does this PR change any of the logging on Tomcat and GAE? How did you verify?

ggalmazor:
That's the beauty of the SLF4J JUL binding. Downstream logging doesn't change at all, just the logging code we write.

I've run the code both on Jetty, Tomcat (on my machine) and GAE and there's no apparent difference

Issue #206, opened 2018-03-08
Sentry integration


ggalmazor:
We can add support for Sentry by following the instructions at: https://docs.sentry.io/clients/java/modules/jul/

Basically, it consists on adding this conf to `logging.properties`:

```properties
# Enable the Console and Sentry handlers
handlers=java.util.logging.ConsoleHandler,io.sentry.jul.SentryHandler

# Override the Sentry handler log level to WARNING
io.sentry.jul.SentryHandler.level=WARNING
```

To configure Sentry (DSN and other settings), probably the most convenient way to do it would be to add a `sentry.properties` file to the classpath.

To make the integration with Sentry useful, we should centralize crash reports of all online Aggregate instances. We need to identify each crash report at least with the Aggregate build version that has produced it, but the following information would also be very useful:
- URL of the git repo (in case it's a fork)
- IP address or FQDN of the Aggregate instance
- Full URL to the Aggregate dashboard page
- Email of the administrator of the Aggregate instance

This information would not only be used to categorize/filter crash reports but also to contact the owners of the instance to ask for collaboration or suggest actions to solve a particular situation.

Some considerations:
- The URL of the git repo can be obtained on build time.
- The IP/FQDN can be obtained at installation time (or run time)
- The Full URL to the Aggregate dashboard page can be ontained at run time.
- How can we obtain the email of the administrator?

ggalmazor:
@opendatakit-bot label "needs discussion"

yanokwa:
Couple of questions come to mind.

Can you confirm that the only change to Aggregate would be to the logging.properties? There'd be no other code change and the other logging (Tomcat, GAE) would work like normal?

What do you mean by the Aggregate dashboard? Is it the GAE dashboard or something else? Why isn't the FQDN enough?

As to the email of the administrator, that's the kind of thing that we would need to ask for in the installer.

ggalmazor:
> Can you confirm that the only change to Aggregate would be to the logging.properties

I'll do a small test to confirm that, but that's what's promised on the Sentry docs...

> What do you mean by the Aggregate dashboard?

You're right, there's no such thing :P I was referring to the submission list (it's the first page you see when accessing Aggregate)

I think we would need this because we don't know where it's really deployed in the IP/FQDN (if it's a Tomcat it can be any path) but I recon it doesn't feel like something critical right now.

> email of the administrator

[...]

PR #211, opened 2018-03-12
WIP - Sentry support

ggalmazor:
Related to #206

In this PR:
- New Sentry dependencies
- A wrapper for SLF4J's LoggerFactory that will initialize Sentry stuff.
- Two patches to be able to run this on GAE:
  - First: a re-implementation of the SentryHandler for JUL that avoids restricted API calls on GAE
  - Second: Sentry is initialized to use a SentryClientFactory that uses GAE's TaskQueue to send crash reports
  - Third: Disable sending crash reports of unhandled exceptions (enabled by default) to avoid restricted API calls on GAE

Pending work:
- All the stuff we do for GAE should be done conditionally depending on where is Aggregate running. This means that the installer must set some property somewhere 
- We might want to ask the user if she wants to send anonymous crash reports to enable or disable Sentry for them.


ggalmazor:
I'm closing this PR since we've changed the monitoring strategy since.

PR #223, opened 2018-03-15
Packer

### LINKED ISSUE #219: Improve UX message on VM startup
I think the /etc/issue should be a little more user-friendly.

```
  > Welcome to ODK Aggregate VM $VERSION
  > 1. Open the web browser on your computer 
  > 2. Go to http://$IP:PORT
  > 3. Sign in with the Aggregate password
  > Need the password? Read the readme.txt file.
  > 
  > If the above URL does not work, these IP addresses instead
  > - 192.168.1.2
```

So the changes here are...
1. Add a bit of indentation (that's what the `  > ` are for) so the message is easier to read.
1. We should drop the HTTPS since that requires configuring the VM with SSL certs. If someone wants to do that, they can probably be figuring things out themselves. 80% of people won't and we don't want to give them a URL that doesn't work. We can do the same for the NAT mapping.
1. Add some $VERSION number so we know what version of the VM is deployed. We can't rely on Aggregate's number because Aggregate might not be launched. 
1. If there are no IP addresses, don't show the IP address message.

### LINKED ISSUE #220: Improve IP address warning for VM
Instead of "Not recommended for production" we should say something like "Not recommended if your IP address can change"

### LINKED ISSUE #221: Change packer variable name
In gradle.properties, I think packerZip or packerUrl is clearer than just packer.

### LINKED ISSUE #222: Remove remote SSH access
`aggregate.json` no longer needs the SSH mapping

ggalmazor:
Closes #219 
Closes #220
Closes #221 
Closes #222 

This PR also improves `issue` message generation hooking it to the DHCP lifecycle, reducing the chances that we won't be able to print any IP address bound to the VM

Manually generated VMs and tested by importing them to VirtualBox
Checked that changing from NAT to bridged mode and back works as expected
Checked that config tool works as expected (setting the FQDN)

Nope

Nope

### DIFF
--- a/packer/ansible/roles/aggregate/files/aggregate-report-ips.sh
+++ b/packer/ansible/roles/aggregate/files/aggregate-report-ips.sh
@@ -5,21 +5,21 @@ PROPS_FILE=/var/lib/tomcat8/webapps/ROOT/WEB-INF/classes/security.properties
 rawHostname=$(grep hostname ${PROPS_FILE} | awk '{split($0,a,"="); print a[2]}')
 hostname=${rawHostname:-localhost}
 httpPort=$(grep port ${PROPS_FILE} | awk '{split($0,a,"="); print a[2]}')
-httpsPort=$(grep securePort ${PROPS_FILE} | awk '{split($0,a,"="); print a[2]}')
+ips=$(ip address | grep "inet " | grep -v "127.0.0.1" | awk '{print $2}' | awk -F'/' '{print $1}')
+VERSION=cocotero
 
-echo "Open a browser and use one of these URLs to access ODK Aggregate"
+echo "> Welcome to ODK Aggregate VM $VERSION"
+echo "> 1. Open the web browser on your computer"
 if [ ${httpPort} = "80" ]; then
-  echo "http://${hostname}"
+  echo "> 2. Go to http://${hostname}"
 else
-  echo "http://${hostname}:${httpPort}"
+  echo "> 2. Go to http://${hostname}:${httpPort}"
 fi
-if [ ${httpsPort} = "443" ]; then
-  echo "https://${hostname}"
-else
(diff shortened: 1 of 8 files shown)


packer/ansible/roles/aggregate/files/aggregate-report-ips.sh
```diff
@@ -1,25 +1,24 @@
...
 
-echo "This VM uses these IP addresses:"
-ip address | grep "inet " | grep -v "127.0.0.1" | awk '{print $2}' | awk -F'/' '{print $1}'
-echo ""
\ No newline at end of file
+if [ ! -z ${ips} ]; then
+  echo ">"
+  echo "> If the above URL does not work, these IP addresses instead"
+  echo "${ips}" | while IFS= read -r line ; do echo "> ${line}"; done
+fi
```
yanokwa: Add an echo ">" after to separate out the message from the login prompt.

packer/ansible/roles/aggregate/files/aggregate-report-ips.sh
```diff
@@ -1,25 +1,24 @@
...
+echo "> 3. Sign in with the Aggregate password"
+echo "> Need the password? Read the readme.txt file."
 
-echo "This VM uses these IP addresses:"
-ip address | grep "inet " | grep -v "127.0.0.1" | awk '{print $2}' | awk -F'/' '{print $1}'
-echo ""
\ No newline at end of file
+if [ ! -z ${ips} ]; then
+  echo ">"
+  echo "> If the above URL does not work, these IP addresses instead"
```
yanokwa: You are missing a try. Maybe 'try connecting to these IP addresses instead'

ggalmazor:
OK, I've amended last commit to commit the last two changes

PR #235, opened 2018-03-22
Add unattended security updates

### LINKED ISSUE #233: Add unattended-upgrades (only for security upgrades) to the VM

yanokwa:
Closes #233

Confirmed that the unattended-upgrades was installed in the ova and updates were not needed. That is, the updates had run.

Followed the instructions at https://wiki.debian.org/UnattendedUpgrades#Unattended_Upgrades

Nothing relly

### DIFF
--- a/packer/ansible/playbook.yml
+++ b/packer/ansible/playbook.yml
@@ -8,3 +8,4 @@
     - tomcat
     - postgresql
     - aggregate
+    - unattended-upgrades
--- a/packer/ansible/roles/unattended-upgrades/tasks/main.yml
+++ b/packer/ansible/roles/unattended-upgrades/tasks/main.yml
@@ -1,0 +1,9 @@
+---
+
+- name: install
+  apt:
+    pkg: unattended-upgrades
+    state: installed
+
+- name: run security-updates
+  command: unattended-upgrades -v


ggalmazor:
I haven't use this before... Reading the docs you linked I found:

> The default configuration file for the unattended-upgrades package is at `/etc/apt/apt.conf.d/50unattended-upgrades`. The defaults will work fine, but you should read it and make changes as needed.
> [...]
> You should at least uncomment the following line:
>     Unattended-Upgrade::Mail "root";

So, have you checked that the defaults work for us? what about uncommenting that line?

Also, the docs suggest two ways of automating the execution of `unattended-upgrade`:
- Automatic call via /etc/apt/apt.conf.d/20auto-upgrades
- Automatic call via /etc/apt/apt.conf.d/02periodic

Which one of these are we adding to the VM?

yanokwa:
Thanks for asking questions!

By default, `etc/apt/apt.conf.d/50unattended-upgrades` and `/etc/apt/apt.conf.d/20auto-upgrades` are both installed. `/02periodic` is not installed. The defaults work for us.

50unattended-upgrades has everything commented out except for the Debian security upgrades. 
```
Unattended-Upgrade::Origins-Pattern {
    "origin=Debian,codename=${distro_codename},label=Debian-Security";
};

20auto-upgrades has this:
```
APT::Periodic::Update-Package-Lists "1";
APT::Periodic::Unattended-Upgrade "1";
```

I did not uncomment `Unattended-Upgrade::Mail "root";` because I don't think the VM should attempt to send email when upgrades happen. Proper email setup adds to the size of the VM (maybe small, though, I didn't check) and I'm pretty sure that without setting up a true mail provider, the messages would be marked as spam. Also, who wants to get an email of updates. The thing should just update, no?

PR #243, opened 2018-05-16
Fix unsafe html strings

ggalmazor:
This PR changes some unsafe uses of HTML and html-capable GWT widgets using SafeHtml instances to ensure dynamic strings are properly escaped.

(GitHub detects changes on whole files for some reason... ignoring whitespaces while reviewing is recommended)

Manually browsed Aggregate screens.

This is just what the GWT recommends when dealing with dynamic html content.

Nope.

N/A

Nope.

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/client/popups/AuditCSVPopup.java
+++ b/src/main/java/org/opendatakit/aggregate/client/popups/AuditCSVPopup.java
@@ -16,6 +16,8 @@
 
 package org.opendatakit.aggregate.client.popups;
 
+import com.google.gwt.safehtml.shared.SafeHtml;
+import com.google.gwt.safehtml.shared.SafeHtmlBuilder;
 import com.google.gwt.user.client.Window;
 import com.google.gwt.user.client.rpc.AsyncCallback;
 import com.google.gwt.user.client.ui.HTML;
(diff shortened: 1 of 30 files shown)


src/main/java/org/opendatakit/aggregate/client/popups/HelpBookPopup.java
```diff
@@ -1,69 +1,70 @@
...
+
+    // populate the panel
+    panel.add(new ClosePopupButton(this));
+
+    BookHelpConsts[] consts = BookHelpConsts.values();
+
+    FlowPanel content = new FlowPanel(); // vertical
+    content.setStylePrimaryName(UIConsts.VERTICAL_FLOW_PANEL_STYLENAME);
+    for (int i = 0; i < consts.length; i++) {
+      // We won't escape these HTML strings using SafeHtmlBuilder because they come from
```
lognaturel: Very zen comment. "they come from...?" 😉

src/main/java/org/opendatakit/aggregate/client/popups/AuditCSVPopup.java
```diff
@@ -99,13 +102,15 @@ static Row from(String csvLine) {
...
-      return "<tr>" +
-          "<td>" + this.event + "</td>" +
-          "<td>" + this.node + "</td>" +
-          "<td>" + this.start.toString() + "</td>" +
-          "<td>" + this.getEnd() + "</td>" +
-          "</tr>";
+    SafeHtml asTr() {
+      return new SafeHtmlBuilder()
+          .appendHtmlConstant("<tr>")
+          .appendHtmlConstant("<td>").appendEscaped(event).appendHtmlConstant("</td>")
```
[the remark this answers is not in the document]
ggalmazor: I'll add it since it will produce the smallest change.

src/main/java/org/opendatakit/aggregate/client/popups/HelpBookPopup.java
ggalmazor: Woops! Fixed :)

ggalmazor:
Thanks, @lognaturel!

PR #245, opened 2018-05-23
Java8 runtime

### LINKED ISSUE #184: Use Java 8 runtime for AppEngine in anticipation of Java 7 runtime deprecation
As discussed in https://forum.opendatakit.org/t/odk-aggregate-deprecation-of-the-java-7-runtime-environment-on-app-engine-standard/11178, Google will deprecate AppEngine Java 7 runtime support on January 16, 2019.

Instructions for migrating: https://cloud.google.com/appengine/docs/standard/java/migrating-to-java8

It would be good to start by confirming whether anything does still require Java 7 or not. I have a vague memory that something might but I'm not sure what it would be.

ggalmazor:
Closes #184 

Built, run and deployed in GAE

This PR just changes version numbers around

Apparently not, but full QA is advised.

Nope.

Nope.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -77,7 +77,7 @@ def versions = [
         mysql     : '5.1.44',
         postgresql: '42.1.4.jre7'
     ],
-    gae           : '1.9.54',
+    gae           : '1.9.63',
     odkTomcatUtil : '1.0.1',
     servletApi    : '3.1.0',
     junit         : '4.12',
--- a/build.gradle
+++ b/build.gradle
@@ -172,7 +172,21 @@ war {
   }
 }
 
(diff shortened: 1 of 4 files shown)


lognaturel:
🎉 My preference would be to merge this and #246 now and then to do a QA pass on master to verify that these two and #243 didn't result in any regressions.

@kkrawczyk123 what do you think? @yanokwa I think this is roughly what you were suggesting in #aggregate-code

yanokwa:
This sounds reasonable to me.

kkrawczyk123:
Sure, but please leave "needs testing" label for me on those pr which you want me to test right after collect's release, it would be easier not to forget about them :)

PR #246, opened 2018-05-23
Update JavaRosa dependency to latest 2.10.0 version

ggalmazor:
This PR is based on #245 and its first commit is https://github.com/opendatakit/aggregate/commit/1e8fbc7235dd85e97b6a6ce5a81bb9c5e0c8b6fd

Compile and run on Tomcat and GAE

After updating the dependency, I had to do the same change as in Briefcase to remove a subclass of `XFormParser` and use the standard implementation from JR.

Nope.

Nope.

Nope.


src/main/java/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
```diff
@@ -222,53 +220,13 @@ public static String xmlWithTimestampComment(String xmlWithoutTimestampComment,
         + xmlWithoutTimestampComment.substring(idx);
   }
 
-  private static class XFormParserWithBindEnhancements extends XFormParser {
```
yanokwa: This change makes me a little nervous and the commit message doesn't explain why this change was made. 

I think this code pulls the `odk:length` attribute from a binding and uses it to set the max length of the DB column. Some of this is explained here: https://opendatakit.org/help/form-design/guidelines/#datastore_string_length

yanokwa: Are you sure JR has support for this attribute now? And can you confirm that forms with length work the same with this change?

lognaturel: @kkrawczyk123 can you please add a test case for this in release testing? `odk:length` is really important because it controls the size of the DB column. You can read more at that link above and let me know if you want me to build you the test form.

kkrawczyk123: Of course!

ggalmazor: I think @yanokwa is right and this change could break this feature. I removed the class because I remembered doing safely the same on Briefcase, but it is after reading it, it's clear that Aggregate depends on reading that attribute to correctly size the storage.

The root reason to remove that class is that it depends on some method in JR that used to be accessible but now it's private. I think I can provide an alternative that doesn't rely on extension to achieve the same result.

yanokwa: Maybe the issue is with MySQL? My guess is that length is used to configure `VARCHAR(n)` in the generation of tables. And performance does change based on this (see https://stackoverflow.com/a/8295254). Would we provide an method in JR for this?

lognaturel: From the `FormDef` it's possible to navigate to the appropriate `TreeElement` and `getBindAttribute("odk", "length")` to read the length bind attribute.

lognaturel: The other thing to look into is whether saving raw bindings might be necessary for checks done when a new version of a form is uploaded.

ggalmazor:
@lognaturel, @yanokwa, I've updated the PR with the suggested changes. Now we do a little post-processing to read any `odk:length` attribute and we still save bindings for later use.

yanokwa:
And can you confirm that you've verified this change prevents truncation on submissions greater than 256 characters if the length is specified?

lognaturel:
Well, I was sure this was going to work great! 😄

PR #254, opened 2018-06-05
Issue 253 new https links param

### LINKED ISSUE #253: Add conf param to enforce HTTPS link generation
#### Problem

As discussed on [this forum post](https://forum.opendatakit.org/t/installing-odk-aggregate-behind-an-ssl-load-balancer-issues/13029/3?u=ggalmazor), and others, some users are trying to add an SSL/TLS layer using a load balancer, instead of configuring Tomcat. This setup is very convenient, especially when dealing with cloud providers like Amazon that make the whole process of adding certificates and setting everything up really easy.

From the point of view of any client trying to speak with the Aggregate server, all communication is secure (HTTPS), but from the point of view of the Aggregate server, everything is done in an open channel (HTTP). Overall, this can be a secure setup, provided that the network has correct security policies in place (this is another whole topic), and it's quite convenient.

Aggregate can be configured to work either in a secure channel (HTTPS) or an open channel (HTTP) and this affects the URLs it builds to let clients know where the form definition files and attached media files can be downloaded from. 

In this scenario, Aggregate should be configured to work on an open channel (HTTP) and this will produce URLs with the HTTP schema which will be rejected by the load balancer.

Some load balancers can be configured to redirect HTTP requests to HTTPS and users can also deploy proxies in their networks to achieve that, but that makes the process of deployment more complex.

#### Proposed solution

Aggregate could have a new configuration parameter to decide whether to use HTTPS or HTTP schema in the links it produces, regardless of the networking configuration.

This would let users configure Aggregate to work on an open channel (HTTP) but generate links that use the HTTPS schema.

ggalmazor:
Closes #253 

- Deployed Aggregate on an AWS EC2 instance and accessed through an Elastic Load Balancer which provided the SSL/TLS layer.
- Tested on an Android device running ODK Collect that I can get a blank form, fill it and send it to Aggregate

This is the narrowest change I could come up with to implement the new conf setting

Nope

Nope

Maybe a doc explaining how to set everything up, but I'd like to have some feedback before going further.


lognaturel:
What's the best approach for getting users to verify this works for them?

yanokwa:
We could probably just generate a WAR...

yanokwa:
@ggalmazor Here's what I'd recommend. Reach out the Kyle at https://forum.opendatakit.org/t/installing-odk-aggregate-behind-an-ssl-load-balancer-issues/13029 and see what is the minimum we could provide (maybe just a WAR) that he could use to try things out. Once we have his go ahead, we can then merge things.

yanokwa:
I'm merging this as part of the code freeze. My hope is that it'll make it easier for Kyle to test.

PR #256, opened 2018-06-12
Issue 255 Fix NPE when uploading a new form version

### LINKED ISSUE #255: NPE when uploading a new version of a form
Aggregate on master commit 6fb527c0d7cea5296eec26608bb5dcd16d901817

#### Problem description
```
java.lang.NullPointerException: null
    at org.opendatakit.aggregate.parser.BaseFormParserForJavaRosa.getBindingsForTreeElement(BaseFormParserForJavaRosa.java:655) ~[classes/:na]
    at org.opendatakit.aggregate.parser.BaseFormParserForJavaRosa.compareTreeElements(BaseFormParserForJavaRosa.java:904) ~[classes/:na]
    at org.opendatakit.aggregate.parser.BaseFormParserForJavaRosa.compareXml(BaseFormParserForJavaRosa.java:821) ~[classes/:na]
    at org.opendatakit.aggregate.parser.FormParserForJavaRosa.guardedInitHelper(FormParserForJavaRosa.java:387) ~[classes/:na]
    at org.opendatakit.aggregate.parser.FormParserForJavaRosa.initHelper(FormParserForJavaRosa.java:236) ~[classes/:na]
    at org.opendatakit.aggregate.parser.FormParserForJavaRosa.<init>(FormParserForJavaRosa.java:148) ~[classes/:na]
    at org.opendatakit.aggregate.servlet.FormUploadServlet.doPost(FormUploadServlet.java:255) ~[classes/:na]
    at javax.servlet.http.HttpServlet.service(HttpServlet.java:707) ~[javax.servlet-api-3.1.0.jar:3.1.0]
    at javax.servlet.http.HttpServlet.service(HttpServlet.java:790) ~[javax.servlet-api-3.1.0.jar:3.1.0]
```
- Upload a form with a version.
- Upload the same form with another version

#### Expected behavior
It should not throw

ggalmazor:
Closes #255 

Uploaded a form and then uploaded the same form with its version changed. 
Verified that the new form gets stored in Aggregate.

Although it's not pretty to parse each XML doc twice, we have no other alternative (yet) because once JR returns a FormDef is too late already and we can't rely on extending the parser as on previous versions of JR.

Nope.

I used this one for my tests: [simple-form.xml.txt](https://github.com/opendatakit/aggregate/files/2095418/simple-form.xml.txt)

Nope.

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
+++ b/src/main/java/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
@@ -418,6 +418,7 @@ public class BaseFormParserForJavaRosa {
     try (StringReader isr = new StringReader(xml)) {
       KXmlParser parser = new KXmlParser();
       parser.setInput(isr);
+      parser.setFeature(XmlPullParser.FEATURE_PROCESS_NAMESPACES, true);
       doc.parse(parser);
     } catch (IOException | XmlPullParserException e) {
       throw new ODKIncompleteSubmissionData(
--- a/src/main/java/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
+++ b/src/main/java/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
@@ -418,10 +418,7 @@ public class BaseFormParserForJavaRosa {
     Document doc = parseXmlToDocument(xml);
 
     List<Element> allBindings = getBindings(doc);
-
-
-    // Copy binding for later use
-    allBindings.forEach(e -> bindElements.add(copyBindingElement(e)));
+    allBindings.forEach(this::storeCopyOfBinding);
 
     // Set lengths of fields if odk:length has been defined
     allBindings.forEach(e -> Optional
@@ -596,6 +593,20 @@ public class BaseFormParserForJavaRosa {
     }
     // clean illegal characters from title
     title = formTitle.replace(BasicConsts.FORWARDSLASH, BasicConsts.EMPTY_STRING);
+  }
+
+  /**
+   * Gets a copy of the given {@link Element} and saves it to be used if
+   * we need to compute the diff between the parsed xml and another xml
+   */
+  private void storeCopyOfBinding(Element binding) {
+    Element copy = new Element();
+    copy.createElement(binding.getNamespace(), binding.getName());
+    for (int i = 0; i < binding.getAttributeCount(); i++) {
+      copy.setAttribute(binding.getAttributeNamespace(i), binding.getAttributeName(i),
+          binding.getAttributeValue(i));
+    }
+    bindElements.add(copy);
   }
 
   private static List<Element> getBindings(Document doc) throws ODKIncompleteSubmissionData {
@@ -1103,18 +1114,6 @@ public class BaseFormParserForJavaRosa {
       if ((retval = element.getAttributeValue(attributeNamespace, attributeName)) != null) {
         return (retval);
       }
-    }
-    return (retval);
-  }
-
-  // copy binding and associated attributes to a new binding element (to help
-  // with maintaining list of original bindings)
-  private static Element copyBindingElement(Element element) {
-    Element retval = new Element();
-    retval.createElement(element.getNamespace(), element.getName());
-    for (int i = 0; i < element.getAttributeCount(); i++) {
-      retval.setAttribute(element.getAttributeNamespace(i), element.getAttributeName(i),
-          element.getAttributeValue(i));
     }
     return (retval);
   }
--- a/src/main/java/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
+++ b/src/main/java/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
@@ -227,7 +227,7 @@ public class BaseFormParserForJavaRosa {
         + xmlWithoutTimestampComment.substring(idx);
   }
 
-  private static synchronized final XFormParser parseFormDefinition(String xml) throws ODKIncompleteSubmissionData {
+  private static synchronized XFormParser parseFormDefinition(String xml) throws ODKIncompleteSubmissionData {
 
     StringReader isr = null;
     try {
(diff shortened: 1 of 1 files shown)


src/main/java/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
```diff
@@ -419,6 +423,17 @@ protected BaseFormParserForJavaRosa(String existingXml, String existingTitle, bo
...
 
+    Document doc = new Document();
+    try (StringReader isr = new StringReader(xml)) {
+      KXmlParser parser = new KXmlParser();
+      parser.setInput(isr);
+      doc.parse(parser);
+    } catch (IOException e) {
+      throw new ODKIncompleteSubmissionData(e, Reason.BAD_JR_PARSE);
+    } catch (XmlPullParserException e) {
+      e.printStackTrace();
```
lognaturel: Log instead?

ggalmazor: I'll fix that, thanks!

ggalmazor:
@lognaturel, I'm sure there's a good reason for everything, but this is quite strange.

I see that the `XFormParse.parseModel(Element e)` actually mutates the input `Document` and removes any `<bind>` and `<itext>` nodes in the `<model>`: :confused:

(line 626)
```java
if (child == null || BIND_ATTR.equals(childName) || "itext".equals(childName)) {
  //Clayton Sims - Jun 17, 2009 - This code is used when the stinginess flag
  //is set for the build. It dynamically wipes out old model nodes once they're
  //used. This is sketchy if anything else plans on touching the nodes.
  //This code can be removed once we're pull-parsing
  e.removeChild(i);
  --i;
}
```

My guess is that if we remove that block it won't harm any other JR process and we will be able to use the complete `Document` after it's parsed.

We would also need to have a getter for the `Document`, which now is `private` inside the `XFormParser`.

lognaturel:
> (line 626)

Aha!! Thanks for identifying that. I believe all parsing is done with kxml2 which is a pull parser. So perhaps a push parser was used at some point and this was for that? It does seem safe to remove though memory usage could go up significantly.

Does that mean that `parser.setFeature(XmlPullParser.FEATURE_PROCESS_NAMESPACES, true);` doesn't actually make a difference?

What I was suggesting is to use the `doc` from `XFormParser.getXMLDocument(isr)` to save the binds and the length attribute. Then later the `XFormParser` object can be built on that doc (what `parseFormDefinition` does currently) and then `parse` (which in this case means transforming the `Document` produced by KXML2 into a `FormDef` -- NOT parsing the XML which was already done by `getXMLDocument`) can be called. It's that last step which loses the binds, right?

ggalmazor:
Hi, @lognaturel!

We have a winner: with the `XmlPullParser.FEATURE_PROCESS_NAMESPACES` feature in the `KXML2Parser` I can get a `Document` and pass it to out `XFormParser` to avoid parsing the XML twice.

I don't really know what enabling that feature does, but now it works as expected. I guess what it does is to make it work :P 

> (One thing that has gotten me tangled up is that there are two meanings of "parse" -- the getXMLDocument method parses the XML into a KXML2 Document which is a tree of nodes. XFormParser.parse() transforms the Document produced by KXML2 into a FormDef. Is that about right? It's the "second parse" which loses the binds.)

It does look like that. I've extracted the two stages into different static methods and I've given them explicit names to clarify it.

src/main/java/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
```diff
@@ -601,28 +579,97 @@ protected BaseFormParserForJavaRosa(String existingXml, String existingTitle, bo
...
+  }
+
+  private static List<Element> getBindings(Document doc) throws ODKIncompleteSubmissionData {
+    // Parse any odk:length in bindings
+    Element head = getChild(doc.getRootElement(), "head");
+    Element model = getChild(head, "model");
+    return getChildren(model, "bind");
+  }
+
+  private static Document parseXmlToDocument(String xml) throws ODKIncompleteSubmissionData {
```
lognaturel: This is now virtually the same as `XFormParser.getXMLDocument` (minus some logging and an unused `CacheTable`). Could you use that one or do you find this makes the exception part clearer, perhaps?

ggalmazor: Thank for spotting that! I didn't realize they were so similar. Anyway, I think I'd prefer not to use `XmlFormParser.getXMLDocument`:
- It's marked as deprecated
- It does text consolidation, which we don't need in Aggregate. We save some cycles there, especially on big forms.
- It has a finer-grained exception handling, which we don't really need either. We wrap the original exception in ours, though. No information is lost.

lognaturel: Wellll, it's not really marked for deprecation. The variant with just a `Reader` param is staying, it's the one with the `CacheTable` that's deprecated. But the other arguments are convincing.

ggalmazor: I've checked the code just in case I missed something... It's a little bit confusing:

```java
public static Document getXMLDocument(Reader reader)throws IOException{
  return getXMLDocument(reader,null);
}

@Deprecated
public static Document getXMLDocument(Reader reader,CacheTable<String> stringCache) throws IOException{
  // Stuff
}
```

It seems like both should have the `@Deprecated` annotation since the first depends on the second. Maybe we should create an issue to fix that on JR...

lognaturel: The one with the reader is needed. It currently calls the more general one to avoid code duplication.

kkrawczyk123:
I've tested this pr on aggregate locally and everything seems to work just fine but as I've never tested any separate pr from aggregate I have to ask if this approach is enough.

kkrawczyk123:

PR #272, opened 2018-07-02
New Docker image build tasks

ggalmazor:
This PR adds new Gradle tasks to build a Docker image with Tomcat and Aggregate. See the [instructions](https://github.com/ggalmazor/aggregate/blob/docker_build_tasks/docs/running-aggregate-with-docker.md) for more info.


docs/running-aggregate-with-docker.md
```diff
@@ -0,0 +1,57 @@
...
+
+  Take note of the tag of the created image. It will appear in a message like this one:
+  
+  ```
+  ...
+  Successfully built bf09fb92a6e9
+  Successfully tagged aggregate:v1.6.0-beta.0-dirty
+  ```
+
+- Launch Aggregate for the first time with `docker run -d -p 8080:8080 -it --name=aggregate aggregate:v1.6.0-beta.0-dirty`.
```
yanokwa: This doesn't seem to work for me. I get the previous message...
```
Successfully built 172dc9613b65
Successfully tagged aggregate:v1.5.0-24-g8bce4afd
```

```
$ docker run -d -p 8080:8080 -it --name=aggregate aggregate:v1.5.0-24-g8bce4afd
7f700ba4b1b1c4e04fe71fd281d1a366cb768ddcdbd0dbaaec7709b4cb4fd2f6
docker: Error response from daemon: OCI runtime create failed: container_linux.go:348: starting container process caused "exec: \"docker-entrypoint.sh\": executable file not found in $PATH": un
```

ggalmazor: That is really strange, because the `Dockerfile` places that file in `/usr/local/bin`, which is in the $PATH.

Could you run `docker -it --entrypoint /bin/bash aggregate:v1.5.0-24-g8bce4afd`? It will start a shell inside the container. Then run `echo $PATH` and copy here the output, please?

yanokwa:
```
$ docker run -it --entrypoint /bin/bash aggregate:v1.5.0-24-g8bce4afd
bash-4.4# echo $PATH
/usr/local/tomcat/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/lib/jvm/java-1.8-openjdk/jre/bin:/usr/lib/jvm/java-1.8-openjdk/bin

Comment (yanokwa) [reply docs/running-aggregate-with-docker.md]: I can also confirm that the file is there

```
bash-4.4# ls -lah /usr/local/bin/docker-entrypoint.sh
-rw-r--r--    1 root     root        2.2K Jul  2 22:49 /usr/local/bin/docker-entrypoint.sh
```

File: docs/running-aggregate-with-docker.md [hunk 2]
```diff
@@ -0,0 +1,57 @@
...
+ 
+  Replace the version with the one you got when building the image
+
+  Access Aggregate at [http:/localhost:8080](http:/localhost:8080)
```
yanokwa: s|http:/|http://

docker/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM tomcat:8.0-jre8-alpine
+RUN rm -rf /usr/local/tomcat/webapps
+COPY webapps /usr/local/tomcat/webapps
+COPY docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
```
yanokwa: After the last copy, you'll probably need this...
```
RUN chmod +x /usr/local/bin/docker-entrypoint.sh
```

docs/running-aggregate-with-docker.md
```diff
@@ -0,0 +1,57 @@
+# Running Aggregate with Docker
+
+## Requirememts
```
yanokwa: s/Requirememts/Requirements/

docs/running-aggregate-with-docker.md
yanokwa: This makes it seem like you'll get Aggregate loading at that URL. You won't right? You'll need to make sure the external DB is configured and working first.

docs/running-aggregate-with-docker.md
```diff
@@ -0,0 +1,57 @@
...
+
+- Install [Docker](https://www.docker.com)
+
+## Quick start - External PostgreSQL instance
```
yanokwa: A little more context here would be useful. Maybe quick instructions on how to connect to a Docker PostgreSQL image?

docker/Dockerfile
ggalmazor: I see in other Dockerfiles that they don't do that. I guess it's because it creates a new hash of the image. In order to ensure this file is executable I've run: `git update-index --chmod=+x docker-entrypoint.sh`

docs/running-aggregate-with-docker.md
ggalmazor: Thanks!

docs/running-aggregate-with-docker.md
ggalmazor: I've added some context and an asciinema playblack with the quick start run.

I would like to avoid having to include a small tutorial on connecting different docker images... 

If a user is already using Docker for PostgreSQL odds are that she already knows about how to link containers or she will prefer taking a look to the docker compose build.

ggalmazor:
I also noticed that we were initializing the database with unencrypted passwords, which is unsecure and unsupported by PostgreSQL from version 10 on.

I've added a commit solving that and, in another commit, I've added another asciinema playback with a run of the quickstart for Docker Compose.

docs/running-aggregate-with-docker-compose.md
```diff
@@ -26,6 +26,8 @@ Run the following commands where you have copied your built setup. **Don't run t
 
 - Stop Aggregate by pressing `ctrl+c`
 
+[![asciicast](https://asciinema.org/a/190218.png)](https://asciinema.org/a/190218)
```
yanokwa: I think asciicasts are cool, but this is the kinda thing that, IMHO, doesn't add a lot and is hard to maintain over time.

docker-compose/initdb/init.sh
```diff
@@ -2,7 +2,7 @@
 set -e
 
 psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" <<-EOSQL
-    CREATE USER "aggregate" WITH UNENCRYPTED PASSWORD 'aggregate';
```
yanokwa: Good catch. There is a similar line in the README. Do we also want that there?

ggalmazor: I'll throw another commit in with that too :)

docs/running-aggregate-with-docker-compose.md
ggalmazor: I think it could relax anyone trying this at home to see how the expected outcome looks like but I guess you're right.

ggalmazor:
I've separated the fixes for `UNENCRYPTED` passwords to let you decide if we merge all or some. Take into account that 7fb9de5dc4f1acb157dab393a1d18578e8b266ac affects the installer, b9d4cc70a5e41361af80ae483506b39089222f69 affects the VM and 9b30923f5981812b910c2acf35894dd5f62e327e affects Docker Compose

I would go for all, but we should re-test those three things to be 100% sure.

yanokwa:
@ggalmazor I think it's fine to merge this, but can you do a test of the various things to make sure it works?

ggalmazor:
Sure!

ggalmazor:
Alright, tried all three and everything is looking fine.

Issue #294, opened 2018-08-08
Add quick primer on encrypted forms to dev docs


lognaturel:
It's easy to forget exactly what info is sent with the encrypted blob and what Aggregate does with it. 

Perhaps https://opendatakit.github.io/xforms-spec/#encryption should also include the submission format.

From @ggalmazor in Slack:

> 1. The public key is attached to the blank form and it is uploaded to Aggregate
>  - Aggregate sees that this is an encrypted form and tricks itself into thinking that the blank form has only three fields: base64EncryptedKey, base64EncryptedElementSignature, and encryptedXmlFile
>  - Here's where the wrong size for those fields happens (<1.5.0 sets that to 2048, 1.6.0 sets that to the default 255)
> 2. Collect pulls the blank form
> 3. The user fills a form and finalizes it
> 4. Collect generates a symmetric key pair (I ignore if it's based on the form's public key or not, but it doesn't matter at this point)
> 5. Collect encrypts the submission into a `submission.xml.enc` file and builds a `submission.xml` file including only three fields:
>  - base64EncryptedKey contains the symmetric key pair encrypted with the form's public key
>  - base64EncryptedElementSignature contains a cryptographic signature of the contents of the submission (using the form's public key again)
>  -  encryptedXmlFile contains the encrypted submission filename 
> 6. Collect sends both files to Aggregate
> 7. Aggregate decomposes the `submission.xml` file and stores the values in the database
>  - It reads 345 chars from base64EncryptedKey but the database truncates that to 255
>  - Same with the signature field
> 8. The user pulls the form with Briefcase
> 9. Sets the export (private key and all) and tries to export it
>  - Briefcase needs to decrypt the symmetric key pair using the private key, but since it was truncated by Aggregate, it's no longer usable
>  - The symmetric key would then be used to decrypt the actual contents of the submission
>  - The signature would also fail decryption for the same reasons. The signature is used to validate that the decrypted contents match the ones the user sent in the first place (they haven't been tampered with)

lognaturel:
I believe that https://opendatakit.github.io/xforms-spec/encryption now has sufficient information. @ggalmazor can you please see if you agree and then close?

ggalmazor:
The docs about encryption in the xforms spec are super complete now. Closing this.

PR #295, opened 2018-08-08
Issue 293 Aggregate sets a wrong size for encryption fields

### LINKED ISSUE #293: Aggregate sets a wrong size for encryption fields
Aggregate v1.16.0

#### Problem description
In #266 we fixed the parsing of the `odk:length` attribute on bindings on non-encrypted forms.

In encrypted forms, we trick Aggregate into thinking that the actual form the user has sent is the one defined in the constant `BaseFormParserForJavaRosa.ENCRYPTED_FORM_DEFINITION`, which is not parsed by the same block that was fixed in #266.

The result is that the `odk:length` attributes of the bindings of `base64EncryptedKey` and `base64EncryptedElementSignature` is ignored and set to the default string length of 255 chars.

**This will truncate the encrypted symmetric encryption keys and the submission cryptographic signature, which require at least 345 chars, __making it impossible to decrypt the submission__**

1. Upload to Aggregate this form
2. Send some submissions with Collect
3. Pull and export the form with Briefcase using this private key

The export will fail

#### Expected behavior
The fields should have a size of 2048 chars, as defined in the constant `BaseFormParserForJavaRosa.ENCRYPTED_FORM_DEFINITION` and exporting the form should work.

#### Other information 
N/A

ggalmazor:
Closes #293

This PR changes the upload form process:
- When Aggregate detects that the form is encrypted, it tricks itself into thinking that the user sent the form defined in `BaseFormParserForJavaRosa.ENCRYPTED_FORM_DEFINITION` which defined field sizes of 2048
- This PR fixes this part of the process to take into account those sizing definitions

- Uploaded [this form](https://github.com/opendatakit/aggregate/files/2270898/encrypted-form.zip)
 and verified with PostgreSQL that the `BASE64_ENCRYPTED_KEY` and `BASE64_ENCRYPTED_ELEMENT_SIGNATURE` fields have a size of 2048 chars
- Uploaded some submissions with Collect
- Pulled the form with Briefcase
- Exported the form with Briefcase

This is the narrowest/quickest patch I came up with reusing what we're already doing with non-encrypted forms

I don't think so. 

(already attached to the verification section)

No.

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
+++ b/src/main/java/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
@@ -542,10 +542,24 @@ public class BaseFormParserForJavaRosa {
       isFieldEncryptedForm = (base64EncryptedFieldRsaPublicKey != null);
     } else {
       isFileEncryptedForm = true;
-      // encrypted -- use the encrypted form template (above) to define
-      // the
-      // storage for this form.
+      // When a form is encrypted, we don't use the actual form that the users sends
+      // because the submissions we will get for this form will only have three fields:
+      // - base64EncryptedKey
+      // - base64EncryptedElementSignature
+      // - encryptedXmlFile
+      //
+      // To address this, we will use the form defined in ENCRYPTED_FORM_DEFINITION instead
+      // which follows the structure described above.
+      // This is discussed in https://github.com/opendatakit/aggregate/issues/294
       XFormParser exfp = parseFormDefinition(ENCRYPTED_FORM_DEFINITION);
+      // Reset bind element and string length maps since we won't be using the original
+      // form parsed and processed at the beginning of this constructor.
+      Document encryptedFormDoc = parseXmlToDocument(ENCRYPTED_FORM_DEFINITION);
+      stringLengths.clear();
+      bindElements.clear();
+      List<Element> encryptedFormBindings = getBindings(encryptedFormDoc);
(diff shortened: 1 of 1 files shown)


ggalmazor:
@opendatakit-bot label "high priority"

src/main/java/org/opendatakit/aggregate/parser/BaseFormParserForJavaRosa.java
```diff
@@ -546,6 +546,13 @@ protected BaseFormParserForJavaRosa(String existingXml, String existingTitle, bo
       // the
       // storage for this form.
       XFormParser exfp = parseFormDefinition(ENCRYPTED_FORM_DEFINITION);
+      // Reset bind element and string length maps since we're actually using another form definition
```
lognaturel: "we're actually using another form definition" -> "we're forcing use of a standard form definition with three fields for all encrypted forms" or something like that? Without the context we discussed on Slack I think it would have taken me a bit to look at that constant and figure out what was going on.

ggalmazor: OK, I'll expand on that

Do you think it would be helpful to add a link to #294?

lognaturel: That sounds like a good idea! Or quickly put together a page to address #294 and link to that.

lognaturel:
Once this is merged I agree that the point release can be done off `master`.

Issue #296, opened 2018-08-08
Specify the correct Java image for CircleCI builds


ggalmazor:
#### Problem description
A recent update on the CircleCI image we use to build JavaRosa has bumped it to use Java 10.0.2 (Oracle), which makes our builds fail.

#### Expected behavior
The build should use Java 8 to work correctly

yanokwa:
Closed by #297

PR #297, opened 2018-08-08
Use latest minor release in CircleCI

ggalmazor:
Addresses #296

Visual inspection.

I like auto-updating to the latest minor release because that's what we expect from our users. It also feels like it gives us early warning of potential problems we'll run into when we upgrade to a major release.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     working_directory: ~/work
     docker:
-      - image: circleci/openjdk:latest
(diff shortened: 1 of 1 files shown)


ggalmazor:
@yanokwa, I tagged you to review this PR so that I can rebase #295 with the fix to the encrypted forms bug we've detected today

ggalmazor:
Thanks, @lognaturel!

lognaturel:
👍 

Would have squashed and merge with message "Use latest openjdk 8 minor release on CircleCI"  :)

PR #300, opened 2018-08-13
Replace System.out.println by logger output in Watchdog (closes #140)

### LINKED ISSUE #140: Reduce watchdog log spamming
Out of box, Aggregate produces 14 lines of log output every 15 minutes. These log entries appear even when nothing out of the ordinary happens, and even when there is no change whatsoever in the state of the application.

My suggestion would be to let the Watchdog report _once_ at INFO level that it is scheduled to do its job every 15 minutes, and then be silent* every time it performs its job as expected. We don't need the application to tell us that it is operating as designed 1344 times a day!

I know that setting `org.opendatakit.aggregate.task.WatchdogWorkerImpl.level=WARNING` in `logging.properties` achieves the same, but I consider that a workaround for what should really be fixed in the implementation.

*) That is, report at FINE level.

zwets:
Closes #140.

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/task/tomcat/WatchdogImpl.java
+++ b/src/main/java/org/opendatakit/aggregate/task/tomcat/WatchdogImpl.java
@@ -259,7 +259,7 @@ public class WatchdogImpl implements Watchdog, SmartLifecycle, InitializingBean,
         logger.debug("Running watchdog task in Tomcat") ;
         impl.checkTasks(cc);
       } catch (Exception e) {
-	logger.error("Exception: {}", e.getMessage());
+        logger.error("Exception: {}", e.getMessage());
         e.printStackTrace();
         // TODO: Problem - decide what to do if an exception occurs
       }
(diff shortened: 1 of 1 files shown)


src/main/java/org/opendatakit/aggregate/task/tomcat/WatchdogImpl.java
```diff
@@ -272,19 +276,19 @@ public void run() {
   public synchronized void createWatchdogTask(long newWatchdogPeriodInMilliseconds) {
     CallingContext cc = new CallingContextImpl();
 
-    logger.info("Changing WatchdogWorker Executor to " + Long.toString(newWatchdogPeriodInMilliseconds) + "ms intervals.");
+    logger.info("Changing WatchdogWorker Executor to {}ms intervals.", newWatchdogPeriodInMilliseconds);
```
ggalmazor: nice!

ggalmazor:
I've merged the PR because the changes only affect to console/log ouptput, but, in a second though, I realize this PR should have followed the standard QA verification process. I'm taking note of this in this comment and tagging @kkrawczyk to test it anyway

ggalmazor:
@opendatakit-bot label "needs testing"

PR #301, opened 2018-08-13
Tidy up

ggalmazor:
This PR is a follow up on @lognaturel's comment at https://github.com/opendatakit/aggregate/pull/295#pullrequestreview-144519396

Uploaded the attached encrypted form and verified in the database that the table fields had the correct size of 2048 chars (almost same process as in #295)

These straightforward changes clarify how the parser deals with field lengths on plain and encrypted forms.

It could be argued that they're really superfluous, but I agree with @lognaturel on that they make the surrounding code more easily understandable.

No.

[This one](https://github.com/opendatakit/aggregate/files/2270898/encrypted-form.zip)

No.


ggalmazor:
Thanks, @dcbriccetti!

PR #303, opened 2018-08-17
Security fixes

ggalmazor:
- Change users from Site Admin > Edit Users
- Delete a form
- Delete submission
- Purge published data
- Purge submissions

First, I've manually verified that all these operations still work in my local Aggregate instance
Second, I've inspected the network interchange of the delete form action and I've resent it using curl from the command line to verify that it won't work since the CSRF token wasn't valid anymore.

This solution implements the [official instructions](http://www.gwtproject.org/doc/latest/DevGuideSecurityRpcXsrf.html) to add CSRF protection to GWT apps.

To reduce code duplication and complexity, I've created a wrapper that handles client-side CSRF requests.

No.

No.

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/client/permissions/AccessConfigurationSheet.java
+++ b/src/main/java/org/opendatakit/aggregate/client/permissions/AccessConfigurationSheet.java
@@ -16,13 +16,33 @@
 
 package org.opendatakit.aggregate.client.permissions;
 
+import com.google.gwt.core.client.GWT;
+import com.google.gwt.event.dom.client.ClickEvent;
+import com.google.gwt.event.logical.shared.ValueChangeEvent;
+import com.google.gwt.safehtml.shared.SafeHtmlBuilder;
+import com.google.gwt.uibinder.client.UiBinder;
+import com.google.gwt.uibinder.client.UiField;
+import com.google.gwt.uibinder.client.UiHandler;
+import com.google.gwt.user.cellview.client.CellTable;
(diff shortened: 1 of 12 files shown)


dcbriccetti:
I expect to look at his today.

dcbriccetti:
So far this looks good, but as someone who is only marginally able to *run* Aggregate (and even then with tons of errors on startup that don’t seem to break the features I need for running Briefcase), I can’t vet this change with confidence in my ability to understand it.

ggalmazor:
Thanks, @dcbriccetti!

src/main/java/org/opendatakit/aggregate/client/permissions/AccessConfigurationSheet.java
```diff
@@ -82,7 +80,9 @@
     userType = new ArrayList<String>();
     userType.add(ACCOUNT_TYPE_ODK);
     userType.add(ACCOUNT_TYPE_GOOGLE);
-  };
+  }
+
+  ;
```
dcbriccetti: By reformatting and exposing these unnecessary semicolons—I count ten—you take on the responsibility of deleting them. 😀

ggalmazor: I wonder how on earth did I miss those! Removing them :)

kkrawczyk123:
I am not sure about purging process so I reported a new issue https://github.com/opendatakit/aggregate/issues/307. Confirm that other process works.

PR #308, opened 2018-08-29
Security fixes batch 2

ggalmazor:
- Export submissions to CSV, JSON, and KML
- Create publishers: Fusion Table, Google Spreadsheet, RedCap Server, SimpleJSON Server, Ohmage Server
- Delete publisher
- Update Publisher API key and restart

~This PR is developed on top of #303 and it starts after commit e3035b0~
(the branch has been rebased after merging #303 into master)

First, I've manually verified that all these operations still work in my local Aggregate instance
Second, I've inspected the network interchange of the delete form action and I've resent it using curl from the command line to verify that it won't work since the CSRF token wasn't valid anymore.

This solution implements the [official instructions](http://www.gwtproject.org/doc/latest/DevGuideSecurityRpcXsrf.html) to add CSRF protection to GWT apps.

To reduce code duplication and complexity, I've created a wrapper that handles client-side CSRF requests.

No.

No.

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/client/externalserv/ServicesAdminService.java
+++ b/src/main/java/org/opendatakit/aggregate/client/externalserv/ServicesAdminService.java
@@ -17,15 +17,13 @@
 package org.opendatakit.aggregate.client.externalserv;
 
 
-import org.opendatakit.aggregate.client.exception.FormNotAvailableException;
+import com.google.gwt.user.client.rpc.RemoteService;
+import com.google.gwt.user.client.rpc.RemoteServiceRelativePath;
 import org.opendatakit.aggregate.client.exception.RequestFailureException;
 import org.opendatakit.aggregate.constants.common.BinaryOption;
 import org.opendatakit.aggregate.constants.common.ExternalServicePublicationOption;
 import org.opendatakit.common.persistence.client.exception.DatastoreFailureException;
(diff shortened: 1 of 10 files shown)


lognaturel:
Apologies for the delay in getting #303 reviewed and merged! This should rebase cleanly on master so we can clearly see the new commits. 🙏

ggalmazor:
I think we're good to go, @lognaturel :)

lognaturel:
Oh, and this isn't super important but for the future, you may want to avoid commit messages like "Tidy up" that don't stand alone. I think they could be squashed with the previous commits they modify or perhaps a name like "Extract repeated callbacks and remove unused ones" would make them more useful on their own.  ¯\_(ツ)_/¯ a mere suggestion!

ggalmazor:
Thanks for the review and comments, @lognaturel!

Yes, normally I'd use better commit messages sorry. That one slipped through, sorry!

I don't like to merge it with another because it makes reviewing it harder, I think.

lognaturel:
> I don't like to merge it with another because it makes reviewing it harder, I think.

Good point, sometimes there's tension between what's useful at review time and what's useful for future `git bisect` or other archaeology. I'm not sure what the best approach in that case is! Probably best to favor review time and help reduce defects in the first place!

PR #318, opened 2018-09-25
Issue 312 fix aggregate config

### LINKED ISSUE #312: aggregate-config only works if ports are 80/433
Aggregate v1.6.1 VM

#### Problem description
If you use run aggregate-config and use a port other than 80 or 443, you will not be able to change passwords. I believe this is the root issue at https://forum.opendatakit.org/t/unable-to-add-or-modify-password-to-users/12624.

* Switch the VM to bridged mode and get an IP (e.g., 192.168.7.61)
* Run the FQDN command `aggregate-config --fqdn 192.168.7.61 --http-port 8080 --https-port 8443`
* Try to connect to the server at 192.168.7.61:8080, it won't work, but 192.168.7.61:80 will. But then you get lots of weird behavior when trying to change the password.

#### Expected behavior
I expect that connecting via 8080 will work. 

#### Other information 
I think the problem here is that Tomcat (server.xml) is listening at port 80/443 and aggregate-config does not change that port, but it should.

ggalmazor:
Closes #312

This PR improves the `aggregate-config` tool that we ship in the Aggregate VM:
- Added a new `--net-mode` argument that will accept either `nat` or `bridge` values to adapt Tomcat's and Aggregate's configuration to the current VM network device mode.

This PR also simplifies the shell command to get the IP addresses on the VM for the MOTD.

Run the commands in the VM, having changed back and forth the network device. Verified that Aggregate was reachable in both nat and bridge scenarios.

No other approaches considered for this change. 

Nope.

No.

https://github.com/opendatakit/docs/issues/866

### DIFF
--- a/packer/ansible/roles/aggregate/files/aggregate-config.sh
+++ b/packer/ansible/roles/aggregate/files/aggregate-config.sh
@@ -3,6 +3,7 @@
 set -o errexit
 
 PROPS_FILE=/var/lib/tomcat8/webapps/ROOT/WEB-INF/classes/security.properties
+TOMCAT_CONFIG_FILE=/var/lib/tomcat8/conf/server.xml
 HELP=NO
 POSITIONAL=()
 while [[ $# -gt 0 ]]
@@ -29,6 +30,11 @@ case $key in
     shift
     shift
     ;;
+    --net-mode)
+    NET_MODE="$2"
+    shift
+    shift
+    ;;
     *)
     POSITIONAL+=("$1")
     shift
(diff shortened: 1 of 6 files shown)


packer/ansible/roles/aggregate/files/aggregate-config.sh
```diff
@@ -49,13 +55,28 @@ showHelp() {
...
   echo "--https-port <value> Set a new HTTPS port"
+  echo "--net-mode   <value> Set the VM's network mode. Use 'nat' or 'bridge'"
 }
 
 if [ ${HELP} = YES ]; then
   showHelp
   exit 0
 fi
 
+if [ ${NET_MODE} != "nat" ] && [ ${NET_MODE} != "bridge" ]; then
```
yanokwa: You'll get a unary operator expected error if you don't have a ${NET_MODE} set. The error will happen on line 66, 73, 111, 116.

Related, you might want to run the script through https://www.shellcheck.net to check for other issues.

packer/ansible/roles/tomcat/files/server.xml
```diff
@@ -1,168 +1,21 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
yanokwa: I typically like keep the stock file and making the necessary changes. I think it makes for easier diffs with the stock file on a default install and the links to the docs are nice too.

But I'm guessing you removed this to make the grepping easier?

packer/ansible/roles/aggregate/files/aggregate-config.sh
```diff
@@ -49,13 +55,28 @@ showHelp() {
...
+  echo "Error: Invalid --net-mode value. Use 'nat' or 'bridge' according to the VM's networking configuration"
+  echo ""
+  showHelp
+  exit 1
+fi
+
+if [ ${NET_MODE} == "bridge" ] && ([ -z ${HTTP_PORT} ] || [ -z ${HTTPS_PORT} ]); then
```
yanokwa: Maybe I'm confused, but this warning is also valid for NAT mode, no?

yanokwa:
Pretty sure this will require a change in the docs. Can you update the PR description?

yanokwa:
Overall, this does what it says on the tin. Just some minor tweaks then it can be merged!

packer/ansible/roles/tomcat/files/server.xml
ggalmazor: That's exactly right. I'll add the stock file as well with a `.original` prefix, just in case

packer/ansible/roles/aggregate/files/aggregate-config.sh
ggalmazor: Not really, but that's mainly because there isn't clear criteria of which args are required which aren't.

I'd propose this:
- If `--http-port` is set, then `--https-port` is required, and vice versa.
- If `--http-port` and `--https-ports` are defined, then `--net-mode` is required.

(FQDN is always optional)

ggalmazor:
OK! Added issue in docs repo and applied your suggestions, @yanokwa.

PR #322, opened 2018-10-09
Security fixes batch 3

ggalmazor:
- Enable accept submissions on form
- Mark submission as complete 
- Enable downloading a form
- Delete exported form
- Delete publisher
- Enable ODK Tables 
- Set ODK App name
- Set skip malformed Submissions

This a follow up on #303 and #308

See #303 and #308

No.

No.

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/client/externalserv/ServicesAdminServiceAsync.java
+++ b/src/main/java/org/opendatakit/aggregate/client/externalserv/ServicesAdminServiceAsync.java
@@ -24,22 +24,15 @@ public interface ServicesAdminServiceAsync {
 
   void getExternalServices(String formid, AsyncCallback<ExternServSummary[]> callback);
 
-  void createFusionTable(String formId, ExternalServicePublicationOption esOption,
-                         String ownerEmail, AsyncCallback<String> callback);
+  void createFusionTable(String formId, ExternalServicePublicationOption esOption, String ownerEmail, AsyncCallback<String> callback);
 
-  void createGoogleSpreadsheet(String formId, String name,
-                               ExternalServicePublicationOption esOption, String ownerEmail, AsyncCallback<String> callback);
+  void createGoogleSpreadsheet(String formId, String name, ExternalServicePublicationOption esOption, String ownerEmail, AsyncCallback<String> callback);
 
-  void createSimpleJsonServer(String formId, String authKey, String url,
-                              ExternalServicePublicationOption es, String ownerEmail, BinaryOption binaryOption,
-                              AsyncCallback<String> callback);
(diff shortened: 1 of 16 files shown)


src/main/java/org/opendatakit/aggregate/client/popups/ConfirmExportDeletePopup.java
```diff
@@ -16,81 +16,70 @@
...
  */
 public final class ConfirmExportDeletePopup extends AbstractPopupBase {
 
-  private static final String BUTTON_ICON = "<img src=\"images/green_right_arrow.png\" />";
+  private static final String BUTTON_ICON = "<img src=\"images/green_right_arrow.png\"/>";
   private static final String TOOLTIP_TXT = "Remove this exported datafile";
-  private static final String HELP_BALLOON_TXT = "This will remove this exported datafile.  You will no longer "
-      + "be able to download this datafile.";
-
-  private final ExportSummary export;
```
lognaturel: Removing these fields is a nice side effect.

src/main/java/org/opendatakit/aggregate/client/externalserv/ServicesAdminServiceAsync.java
```diff
@@ -24,22 +24,15 @@
 
   void getExternalServices(String formid, AsyncCallback<ExternServSummary[]> callback);
 
-  void createFusionTable(String formId, ExternalServicePublicationOption esOption,
-                         String ownerEmail, AsyncCallback<String> callback);
+  void createFusionTable(String formId, ExternalServicePublicationOption esOption, String ownerEmail, AsyncCallback<String> callback);
```
lognaturel: I dislike some of these newline changes because I often have two windows side by side or review on Github and now I have to scroll horizontally. I still prefer an 80-char width limit. But I won't push back too hard. Was this an automatic IntelliJ change?

lognaturel:
Also, you're so fast, @ggalmazor! 🏃

src/main/java/org/opendatakit/aggregate/client/popups/ConfirmExportDeletePopup.java
ggalmazor: Yup! Clean up season! :)

src/main/java/org/opendatakit/aggregate/client/externalserv/ServicesAdminServiceAsync.java
lognaturel: > way easier to read and change them when all members are displayed sequentially

That is very interesting! Even when writing and modifying code, I prefer shorter lines. But I can see your point. Anyway, not a big deal.

PR #323, opened 2018-10-09
Add note warning about a new Aggregate version

ggalmazor:
This PR attempts to give users a heads up about the availability of a newer Aggregate version.

A new note will appear on the Filters tab at the bottom left indicating the current version and a link if there's a newer version available. Otherwise, it tells the user that they're "up to date".

image

Run on my dev server and verified that both messages where correctly being shown.

It's a very narrow implementation, trying to have minimal impact on the layout

It's, by the way, the most cross-browser, backwards-compatible JS code I've written in my life:

```js
var req = new XMLHttpRequest();
req.open('GET', 'https://api.github.com/repos/opendatakit/aggregate/releases/latest', false);
req.send(null);
if (req.readyState === 4 && req.status === 200)
  return JSON.parse(req.responseText).tag_name;
```
(Notice that old-school `XMLHttpRequest`! It's even synchronous!)
Nope.

None.

No.

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/client/FiltersDataPanel.java
+++ b/src/main/java/org/opendatakit/aggregate/client/FiltersDataPanel.java
@@ -16,7 +16,8 @@
 
 package org.opendatakit.aggregate.client;
 
-import com.google.gwt.dom.client.Style;
+import static org.opendatakit.aggregate.client.LayoutUtils.*;
+
 import com.google.gwt.user.client.ui.FlexTable;
 import com.google.gwt.user.client.ui.FlowPanel;
 import com.google.gwt.user.client.ui.HTML;
@@ -28,7 +29,6 @@ import com.google.gwt.user.client.ui.VerticalPanel;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Map;
-import org.opendatakit.aggregate.buildconfig.BuildConfig;
 import org.opendatakit.aggregate.client.filter.ColumnFilter;
 import org.opendatakit.aggregate.client.filter.Filter;
 import org.opendatakit.aggregate.client.filter.FilterGroup;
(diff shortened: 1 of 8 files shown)


src/main/java/org/opendatakit/aggregate/client/FiltersDataPanel.java
```diff
@@ -91,12 +90,35 @@ public FiltersDataPanel(FilterSubTab parentPanel) {
     filtersTree = new Tree();
     panel.add(filtersTree);
 
+    String latestVersion = getLatestVersion();
+    boolean needsUpdate = !BuildConfig.VERSION.startsWith(latestVersion);
```
lognaturel: Is this `startsWith` instead of `equals` so that someone running a local version with a commit hash isn't told to upgrade?

lognaturel:
I bet @yanokwa will have some visual tweaks to suggest but I'm very happy with this.

> the most cross-browser, backwards-compatible JS code I've written in my life
😄🏆

src/main/java/org/opendatakit/aggregate/client/FiltersDataPanel.java
ggalmazor: Yes, that's right

yanokwa:
@ggalmazor Couple of questions...
1. Is it possible to put this on every page as a footer? That feel better to me if it's not crazy complicated to add.
1. Seems like we should also put it in the version information in prefs since that's where version information is.

I'm thinking we can make this shorter so it'll fit everywhere.
* v1.6.1 - You're up to date!
* v1.6.1-asdasd-dirty - [Update available](http://example.com)

lognaturel:
> Is it possible to put this on every page as a footer? That feel better to me if it's not crazy complicated to add.

How about filing this as a help wanted issue? I believe that if @ggalmazor didn't just do that, it's because adding a footer is not entirely trivial. That's expected because there is no footer currently. Doesn't feel worth investing in right now. 

> Seems like we should also put it in the version information in prefs since that's where version information is.

That sounds good and like it's probably not too big a lift.

Putting everything on the same line sounds good to me. It may wrap for people with local, unreleased builds but it doesn't really matter since updates aren't really relevant for them anyway.

@yanokwa, note that `v1.6.1-asdasd-dirty` would not tell the user to update if the latest version is `v1.6.1` but that `v1.4.1-asdasd-dirty` would. I think that's the right decision.

ggalmazor:
It turns out that the `native` method magic works just like normal methods in Java and I've been able to extract & reuse the notes stuff.

I've added the version notes to all sub-tabs in Submissions and Form Management tabs.

Tradeoff: these notes have fixed positions, which means that they aren't affected by the flow of UI elements, which means that they will get on top of other things when normal content reaches the bottom of the page.

Let me know what you think ;)

yanokwa:
I attached a screenshot of the overflow to save someone running it. Possible to make that background white so it's at least legible?
screen shot 2018-10-11 at 09 13 32

Why not also add it to Site Admin (just curious).

ggalmazor:
I've given the note a white (.9 opacity) background and I've added it to the Site Admin sub tabs (it's not super terrible when it overlaps).

kkrawczyk123:
Improved and tested in https://github.com/opendatakit/aggregate/pull/328

PR #325, opened 2018-10-10
Use language from the UI and provide more actionable information on password

### LINKED ISSUE #324: Aggregate password warning is not scary enough
Related to https://github.com/opendatakit/docs/pull/872

When you have a server with a default password, the warning you get is:
> This server and its data are not secure! Please change the super-user's password!

This is a little vague and doesn't capture the damage that could occur. I propose we change it to.
> Your super-user account is using the default password. Please change the password immediately to prevent authorized users from taking control of your server!

yanokwa:
Closes #324 

Ran the build locally to ensure it loaded in Chrome and Firefox. Also made sure the text wraps nicely.

I tried adding HTML tags and Java newlines in hopes to improving readability. Those show up unprocessed as `<a href` and `\n`.

No. It's a very targeted change.

Yes, once this is merged, the copy at https://docs.opendatakit.org/aggregate-admin/#permissions needs to be changed.


kkrawczyk123:
**Tested with success!**
Checked out on Chrome, Firefox and Safari. Much scarier than it was before!

Issue #290, opened 2018-07-20
Deprecate publishers we can't verify as functional


yanokwa:
I'd like to deprecate the following publishers:

* RedCap Server because it's a closed community where we can't even ask if the publisher is used. Also, according to https://forum.opendatakit.org/t/redcap-to-odk/1477/30 I'm pretty sure it doesn't work.
* Ohmage JSON server because according to https://github.com/opendatakit/opendatakit/wiki/Aggregate-Publishers-Implementation-Details#known-issues it has no verified accepting servers. We shouldn't ship things we can't verify as working. I'd be surprised if anyone is using it.

My proposed approach here is pretty direct.
* In the next release, ~~replace Z-ALPHA in the type and~~ prepend (Deprecated). This will allow users who have those publishers to see their status in the publish status page.
* ~~Remove the choice from the drop down list so users cannot create new publishers.~~
* Update the docs to reflect this change.

Note that we want people to be able to delete publishers even after we remove the choice.

I could be convinced that we stagger this over two or three releases if everyone thinks my direct approach is unkind to users. But you know we can point out this deprecation in the beta and people don't have to upgrade.

aurdipas:
I agree on your approach. +1

smoyth:
Works for me. A message on the forum in advance might also be a nice courtesy.

PR #326, opened 2018-10-10
Warn users of upcoming deprecation

yanokwa:
Address #290

Ran the app and confirmed the publish dropdown is there in Submissions tab. Also confirmed that publisher types in Published Data showed up.

I did not try this with an install with an existing publisher so it'd be good to test that users that upgrade won't be harmed.

Just a targeted text change.

There might be people who use the text of this drop down to do things, but given that these publishers don't work, it's unlikely.

Not yet. After we deprecate in v2.0 then we should update the docs.


kkrawczyk123:
**Tested with success!**
Information about deprecation is displayed next to both servers on the publish list on the local instance and also after upgrading GAE instance.

PR #328, opened 2018-10-17
Small UI improvements to the version note

### LINKED ISSUE #330: Version information makes it difficult to download form XML
1.7.0-beta.0

#### Problem description
If you have a long list of forms, the version number overlaps so you can't click on the form to download it. 

screen shot 2018-10-18 at 21 44 53

#### Potential solution
Maybe we move the version to next to the Site Admin tab?
screen shot 2018-10-18 at 21 47 51
And since this is more visible, maybe we only show the version number when there is no update available (drop the "You're up to date!").

Alternatively, can we add some whitespace to the end of the form listing? The site preferences seems to do something like that.

ggalmazor:
Closes #330 

This PR adds a background to the help layer to hide the version note

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/client/AggregateUI.java
+++ b/src/main/java/org/opendatakit/aggregate/client/AggregateUI.java
@@ -16,6 +16,7 @@
 
 package org.opendatakit.aggregate.client;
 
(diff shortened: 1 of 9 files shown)


kkrawczyk123:
**Tested with success!**

PR #333, opened 2018-11-08
Issue 287 remove odk2 code

ggalmazor:
Closes #287

Try this PR:
- [Tomcat WAR file](https://drive.google.com/open?id=1Tey-AaGrjYl5P7Ka4XgHBEcUsAwBcdb1)
- [Installer - Linux](https://drive.google.com/open?id=1Qf2etYstCAHBFBiAUnDYVcPTh4sUq6S1)
- [Installer - Linux 64 bits](https://drive.google.com/open?id=1WpGDrRNGxTFoEx6zVubYBdYvytbm0k3h)
- [Installer - macOS](https://drive.google.com/open?id=1DfyFrmPVUsxwWGnAOjHUQWkg_BdBPaxX)
- [Installer - Windows](https://drive.google.com/open?id=1Ihm82lLHYiXayxKyTjYqINdqrlPLNthX)

#### What has been done to verify that this works as intended?

Run Aggregate on my machine, and deployed it to https://aggregate-gigi.appspot.com

#### Why is this the best possible solution? Were any other approaches considered?
N/A

#### Are there any risks to merging this code? If so, what are they?

Users should expect no more ODK Tables related stuff. This includes:

- No servlets for third-party web clients that were using them, including:
  - `/aggregateui/odktables/*`
  - `/aggregateui/odktablesadmin`
  - `/aggregateui/serverdataservice`
  - `/aggregateui/serverdiffservice`
  - `/aggregateui/servertableservice`
  - `/aggregateui/servertableaclservice`
- Removed options from the Preferences tab:
  - Field to set the app name
  - Checkbox to enable ODK Tables
- Removed description of grants related to ODK Tables from the Permissions tab

This PR has very invasive changes across all aspects of Aggregate. Careful QA should be performed.

#### Do we need any specific form for testing your changes? If so, please attach one
No

#### Does this change require updates to documentation? If so, please file an issue at https://github.com/opendatakit/docs/issues/new and include the link below.
I guess it does require changes to documentation. I'll add a link here after getting some feedback.

PR #335, opened 2018-11-14
Issue 334 code cleanup

### LINKED ISSUE #334: Code cleanup: reformat, remove unused code, etc.
After working on #287, it looked like we should do some extra cleaning up:

- Reformat code
- Run simple inspections and:
  - Remove redundant throws clauses
  - Remove unused code
  - Fix dangling, problematic, confusing JavaDocs

ggalmazor:
Closes #334

Run Aggregate and test manually that generally everything works: uploading, exporting, publishing forms, credentials management, etc.

The changes on this PR where driven by IntelliJ's code inspections, targetting unused artifacts, thrown exceptions and JavaDocs mainly.

I've included JavaDocs in this run because `{@link}` and `@throws` clauses where marking the related classes as being used even if the code didn't actually use/throw them.

Yes, although no changes in behavior should be seen. I could have broken something. Manual thorough testing is required.

Nope.

No.


dcbriccetti:
752 files changed‽ I didn’t know there were that many files in all of ODK! :-)
I’ll do some spot checking.

ggalmazor:
Thanks, @dcbriccetti!

I think going one commit at a time will make the review easier for you, since the commits targetting bulk safe-ish refactors (where all those changed files come from) are different than others with more important and riskier changes.

Also, The review order should go #333, then #335, then #338

Issue #360, opened 2018-12-26
Can't filter by temporal metadata fields


ggalmazor:
#### Software and hardware versions 
Aggregate v2.x

#### Problem description
After the ISO 8601 adaptations, where temporal fields started to behave like GEOPOINT fields (data persisted as multiple columns), these filters no longer work, since they don't have a backing data model.

#### Steps to reproduce the problem
Filter by submission date or "date marked as complete"

#### Expected behavior
Filtering submissions with these fields should work like any other temporal fields.

#### Other information 
QueryBase.addFilterChildren() fails to `attr.getFormDataModel().getChildren()` because `attr.getFormDataModel()` is null.

ggalmazor:
Closed by #365

PR #365, opened 2018-12-28
Issue 359 filter metadata date marked as complete

### LINKED ISSUE #359: Can't filter by metadata field date_marked_as_complete
Aggregate v1.?.? (since 2011)

#### Problem description
The server fails to filter submissions by the metadata field "date_marked_as_complete".

Show metadata fields in the submissions list and try to add a filter using this field.

#### Expected behavior
It should work

#### Other information 
There's a problem with a `switch` block in `QueryBase`. The `case` for this field is missing the `break`;

ggalmazor:
Closes #359, and #360

Try to filter submissions by submission date or "date marked as complete" metadata fields and verify it works

This is the narrowest solution I could thin of.

Nope.

No.

No.

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/query/submission/QueryByUIFilterGroup.java
+++ b/src/main/java/org/opendatakit/aggregate/query/submission/QueryByUIFilterGroup.java
@@ -38,7 +38,6 @@ import org.opendatakit.aggregate.server.ServerPreferencesProperties;
 import org.opendatakit.aggregate.server.UITrans;
 import org.opendatakit.aggregate.submission.Submission;
 import org.opendatakit.aggregate.submission.SubmissionKey;
-import org.opendatakit.aggregate.submission.type.jr.JRTemporal;
 import org.opendatakit.common.datamodel.ODKEnumeratedElementException;
 import org.opendatakit.common.persistence.CommonFieldsBase;
 import org.opendatakit.common.persistence.Query;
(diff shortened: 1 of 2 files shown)


src/main/java/org/opendatakit/aggregate/query/submission/QueryByUIFilterGroup.java
```diff
@@ -135,50 +135,72 @@ public QueryByUIFilterGroup(IForm form, FilterGroup filterGroup, CompletionFlag
...
-            continue;
+            break;
           case JRTIME:
-            compareValue = WebUtils.parseDate(value);
             super.addFilterChildren(fem, column.getChildColumnCode(), op, compareValue);
-            continue;
+            break;
           case INTEGER:
-            compareValue = Long.valueOf(value);
+            super.addFilter(fem, op, compareValue);
```
dcbriccetti: How about grouping together these four cases so this only appears once:
`super.addFilter(fem, op, compareValue);`

src/main/java/org/opendatakit/aggregate/query/submission/QueryByUIFilterGroup.java
```diff
@@ -135,50 +135,72 @@ public QueryByUIFilterGroup(IForm form, FilterGroup filterGroup, CompletionFlag
...
       }
     }
 
   }
 
+  public Object getCompareValue(FormElementModel fem, String value) {
+    switch (fem.getElementType()) {
+      case BOOLEAN:
+        return WebUtils.parseBoolean(value);
+      case JRDATETIME:
```
dcbriccetti: Same suggestion here about consolidating like cases

src/main/java/org/opendatakit/aggregate/query/submission/QueryByUIFilterGroup.java
```diff
@@ -139,6 +139,10 @@ public QueryByUIFilterGroup(IForm form, FilterGroup filterGroup, CompletionFlag
 
         switch (fem.getElementType()) {
           case BOOLEAN:
+          case INTEGER:
+          case DECIMAL:
+          case SELECT1:
+          case STRING:
```
dcbriccetti: Much cleaner!

ggalmazor: Yup, you have an eye for this :) Thanks!

PR #372, opened 2019-01-09
WIP - CloudConfig setups and tools for DigitalOcean and local (VirtualBox)

### LINKED ISSUE #356: Create instructions and artifacts (Packer) for transitioning to DigitalOcean
To do:
- [ ] Packer build
- [ ] Docs

Features:
- [ ] SSL support
- [ ] Upgrade system
- [ ] Network configuration tool

ggalmazor:
Closes #356

See https://github.com/ggalmazor/aggregate/tree/cloud-config/cloud-config
- [Digital Ocean instructions](https://github.com/ggalmazor/aggregate/blob/cloud-config/cloud-config/digital-ocean)
- [Local (VirtualBox) instructions](https://github.com/ggalmazor/aggregate/blob/cloud-config/cloud-config/local-virtualbox)


ggalmazor:
@yanokwa this is a stable version with almost all the pieces we're going to need. I'll send in another one so that we can explore the issue with the apt Tomcat server

@kkrawczyk123, maybe you can try the instructions to deploy locally

ggalmazor:
All the action points have been addressed and this PR is ready for review now.

ggalmazor:
After researching and trying out Floating IPs I think they don't really offer an advantage for the average user. Unless we can find a reason to add extra info about it, I've decided not to do so.

Floating IPs can only be created for running Droplets. This doesn't let you avoid the awkward waiting time between you get your Droplet's public IP, you set your domain up, and the domain change propagates.

I can only think tha Floating IPs could offer an advantage for users wanting to scale up/down their Droplets, since they could do this with virtually zero downtime but most of the time users don't need a zero downtime transition and it would require database replication before attempting a Droplet replacement, which can be hard for users to set up.

Users wanting to have lots of horizontal scalability could also combine Floating IPs and a Load Balancer, but our current CloudConfig setup doesn't support that.

PR #374, opened 2019-01-11
Issue 373 Temporal filters after upgrading to v2.0

### LINKED ISSUE #373: Can't filter by temporal fields after upgrading to v2 on pre-existing forms
Aggregate v2.0

#### Problem description
Adding a filter using a date field has no effect.

- Load this database dump, corresponding to a baseline Aggregate v1.7.1, which includes a form with temporal fields and one submission: [odk_v1.sql.zip](https://github.com/opendatakit/aggregate/files/2748511/odk_v1.sql.zip)

  ```bash
  unzip odk_v1.sql.zip
  psql -h localhost -f odk_v1.sql odk
  ```

  Then change the `jdbc.properties` file to use the `odk_v1` schema
  (using date and dateTime filters in an Aggregate v1.7.1 running on this database will work)

- Launch Aggregate v2.x
- Use any of these filters:
  - `some-date <= 2019-01-01`
  - `some-time <= 00:00:00.000Z`
  - `some-datetime <= 2019-01-01T00:00:00.000Z`

#### Expected behavior
You should see no submissions, because the only uploaded submission has 2019 date and dateTime values.

#### Other information 
N/A

ggalmazor:
Closes #373

Manually tested following the instructions on #373.

This PR improves the detection of temporal fields of forms uploaded before upgrading to v2.0
- They can lack their FormElementModel
- They can be metadata fields
- They can have a FormElementModel without child elements

Nope.

Yes. The instructions to reproduce the scenario are attached to #373

Nope.

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/query/submission/QueryByUIFilterGroup.java
+++ b/src/main/java/org/opendatakit/aggregate/query/submission/QueryByUIFilterGroup.java
@@ -148,15 +148,13 @@ public class QueryByUIFilterGroup extends QueryBase {
             super.addFilter(fem, op, compareValue);
             break;
           case JRDATETIME:
-            if (fem.getFormDataModel() == null && fem.getElementName().contains("meta-"))
-              super.addFilter(fem, op, compareValue);
-            else if (fem.getFormDataModel() != null)
-              super.addFilterChildren(fem, column.getChildColumnCode(), op, compareValue);
-            else
-              throw new IllegalArgumentException("Unrecognized dateTime field");
-            break;
           case JRDATE:
           case JRTIME:
+            if (isLegacyTemporalField(fem))
+              super.addFilter(fem, op, compareValue);
+            else
+              super.addFilterChildren(fem, column.getChildColumnCode(), op, compareValue);
+            break;
           case GEOPOINT:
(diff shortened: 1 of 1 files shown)


src/main/java/org/opendatakit/aggregate/query/submission/QueryByUIFilterGroup.java
```diff
@@ -148,15 +148,13 @@ public QueryByUIFilterGroup(IForm form, FilterGroup filterGroup, CompletionFlag
...
-            if (fem.getFormDataModel() == null && fem.getElementName().contains("meta-"))
+          case JRDATE:
+          case JRTIME:
+            if (isLegacyTemporalField(fem))
               super.addFilter(fem, op, compareValue);
-            else if (fem.getFormDataModel() != null)
-              super.addFilterChildren(fem, column.getChildColumnCode(), op, compareValue);
             else
-              throw new IllegalArgumentException("Unrecognized dateTime field");
+              super.addFilterChildren(fem, column.getChildColumnCode(), op, compareValue);
```
ggalmazor: I've removed the exception because now I know more about this and I'm more confident about the range of scenarios we could get.

PR #382, opened 2019-01-16
Issue 380 Fix geopoint formatting when it lacks alt or acc components

### LINKED ISSUE #380: Publish to Google Spreadsheet is not completed.
Windows/Linux/macOS, Aggregate v1.2.x, Java vx.x.x, ...

#### Problem description
When Publish submissions to Google Spreadsheet I am obtaining only column names in the sheet, but submissions are not visible (All widgets form) and ex. when tried to publish Birds form I have only one 1 submission in the sheet visible. 

Publish to Google Spreadsheet, upload existing only. Open the sheet. 

#### Expected behavior
All submissions should be successfully published and visible in the created sheet.

ggalmazor:
Closes #380

Added new test cases to verify that we can format geopoints that lack altutide or accuracy components

This is the smallest change I could come up that takes into account that a geopoint value, and its lat, lon, alt, and acc can be `null`.

Nope.

No. Any form with geopoints will do.

No.

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/format/element/BasicElementFormatter.java
+++ b/src/main/java/org/opendatakit/aggregate/format/element/BasicElementFormatter.java
@@ -116,25 +116,30 @@ public class BasicElementFormatter implements ElementFormatter {
       return;
     }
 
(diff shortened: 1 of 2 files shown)


kkrawczyk123:
**Tested with success!**
Export to csv confirmed as well.

@opendatakit-bot "label "behavior verified"

PR #403, opened 2019-02-05
Issue 401 check hostnames conditionally

### LINKED ISSUE #401: Inifinite redirect loop behind proxy/load balancer
Linux(Ubuntu 18.04), Aggregate v2.0.0

#### Problem description

When Aggregate is running behind a proxy, the detected/configured hostname won't match the HTTP request's hostname, which will trigger a 302 response, starting an infinite redirection loop.

Run the Cloud-Config stack and change the security.properties file (set the hostname to the IP or domain name that final users will see) and try to access Aggregate

#### Expected behavior

It should respect whatever it's configured as long as requests reach the server and not redirect

#### Other information 
The redirection might be useful in other scenarios (undetermined at this point). We should add a new configuration param to toggle this behavior.

The Aggregate-CLI tool must be updated https://github.com/opendatakit/aggregate-cli/issues/7

ggalmazor:
Closes #401

Also fixes the Cloud-Config stack to install only releases (remove the `--include-pre-releases` flag from the aggregate-cli install command)

Run locally with the Cloud-Config stack for VirtualBox
Upload a [Simple Form](https://github.com/opendatakit/briefcase/blob/master/test/resources/org/opendatakit/briefcase/export/simple-form.xml), a [Media Files](https://github.com/opendatakit/briefcase/blob/master/test/resources/org/opendatakit/briefcase/export/media-files.xml) form, and [Birds](https://github.com/opendatakit/sample-forms/archive/master.zip)
Get the forms with my android device, fill some subs and send them
Pull the forms with Briefcase and check the files in my storage directory

I've also tried it in DigitalOcean using the latest aggregate-cli and emulating a release in my fork.

This is the simplest solution I can think of to conditionally avoid redirecting to the canonical hostname (either detected on startup or configured via security.properties file) when the HTTP request's hostname doesn't match it.

No.

The Aggregate-CLI needs to be updated with this
https://github.com/opendatakit/aggregate-cli/issues/7

Any form will do, but I've linked the ones I've used above.

No.

### DIFF
--- a/cloud-config/assets/cloud-config.yml
+++ b/cloud-config/assets/cloud-config.yml
@@ -86,6 +86,6 @@ runcmd:
   - su postgres -c "psql -c \"CREATE SCHEMA aggregate\" aggregate"
   - su postgres -c "psql -c \"GRANT ALL PRIVILEGES ON SCHEMA aggregate TO aggregate\" aggregate"
 
-  - aggregate-cli -i -ip -y -c /root/aggregate-config.json
+  - aggregate-cli -i -y -c /root/aggregate-config.json
 
   - service nginx restart
--- a/cloud-config/assets/cloud-config.yml
+++ b/cloud-config/assets/cloud-config.yml
@@ -33,7 +33,8 @@ write_files:
         "security": {
           "forceHttpsLinks": true,
           "port": 80,
-          "securePort": 443
+          "securePort": 443,
+          "checkHostnames": false
         },
         "tomcat": {
           "uid": "tomcat8",
--- a/cloud-config/assets/cloud-config.yml
+++ b/cloud-config/assets/cloud-config.yml
@@ -31,6 +31,7 @@ write_files:
           "password": "aggregate"
         },
         "security": {
+          "hostname": "foo.bar",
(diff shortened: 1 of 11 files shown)


docs/aggregate-config.md
```diff
@@ -68,6 +68,15 @@ This file supports the following configuration settings:
   
   Set this to the security realm you want your users to see when logging into Aggregate   
 
+**security.server.checkHostnames*
```
yanokwa: Missing a `*` at the end of this and forceHttpsLinks

cloud-config/assets/cloud-config.yml
```diff
@@ -86,6 +86,6 @@ runcmd:
   - su postgres -c "psql -c \"CREATE SCHEMA aggregate\" aggregate"
   - su postgres -c "psql -c \"GRANT ALL PRIVILEGES ON SCHEMA aggregate TO aggregate\" aggregate"
 
-  - aggregate-cli -i -ip -y -c /root/aggregate-config.json
```
yanokwa: Good catch! I think we had this in to do testing.

PR #406, opened 2019-02-06
Fix sed error that breaks installs

yanokwa:
Confirmed this works with clean install on DO. Also confirmed HTTPS certs are installed.

Narrowest fix possible. 

Yes. I am still getting infinite redirects even after using aggregate-cli to install a custom war. Not sure where the problem is, but my recommendation is that we get as close to a stock install as possible. So I'd recommend... 
1. Merge this change.
1. Publish master as a temporary v2.0.0 pre-release WAR on GitHub
1. Copy this PRs [cloud-config](https://raw.githubusercontent.com/opendatakit/aggregate/bf0d82d7fde05d4bc1417f1bf77a2958db29b901/cloud-config/assets/cloud-config.yml) into DO.
1. Edit the config so aggregate-cli gets pre-releases (add `-ip`)
1. Confirm that cloud-init-output is clean
1. Run certbot to get HTTPS working
1. Confirm if page loads with no redirects

No.

### DIFF
--- a/cloud-config/assets/cloud-config.yml
+++ b/cloud-config/assets/cloud-config.yml
@@ -86,8 +86,8 @@ runcmd:
   - su postgres -c "psql -c \"CREATE SCHEMA aggregate\" aggregate"
   - su postgres -c "psql -c \"GRANT ALL PRIVILEGES ON SCHEMA aggregate TO aggregate\" aggregate"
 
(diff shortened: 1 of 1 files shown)


ggalmazor:
I think you're getting the redirects because you haven't refreshed your `security.properties` and `odk-settings.xml` files with the new example files that include the new conf params.

I'm double checking it anyway.

PR #407, opened 2019-02-06
v2.0.0 release tweaks

### LINKED ISSUE #411: Force package upgrades in Cloud-Config stacks
The Cloud-Config stacks report on login:

```
45 packages can be updated.
11 updates are security updates.
```

Should we take care of upgrading the packages after completing the installation so that those security updates get installed?

yanokwa:
Closes #411

* Checked to make sure this repo and the CLI repo don't call tar, vim, or tree.
* Checked on VPS to make sure `curl -s` doesn't show progress bar
 
### DIFF
--- a/cloud-config/assets/cloud-config.yml
+++ b/cloud-config/assets/cloud-config.yml
@@ -57,7 +57,7 @@ write_files:
     permissions: '0755'
     content: |
       #!/bin/sh
-      curl -s https://api.github.com/repos/opendatakit/aggregate-cli/releases/latest \
+      curl -sS https://api.github.com/repos/opendatakit/aggregate-cli/releases/latest \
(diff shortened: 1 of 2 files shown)


cloud-config/assets/cloud-config.yml
```diff
@@ -67,6 +67,8 @@ runcmd:
   - download-aggregate-cli
 
   - apt-get -y remove openjdk-11-jre-headless
+  - unattended-upgrades
```
ggalmazor: Does this package come pre-installed in Ubuntu 18.04?

yanokwa:
I checked and it does, but the whole PR should be verified.

ggalmazor:
OK!

Issue #209, opened 2018-03-09
User can "publish" submissions/forms to servers without choosing any fields,


kkrawczyk123:
#### Software and hardware versions 
Aggregate v1.4.x, Java v1.8.x, ...

#### Problem description
User is able to publish data to servers without choosing any fields. I think that some information about required fields should be added or publish button may be disabled or error message should inform user what went wrong and give a chance to correct users choices.

#### Steps to reproduce the problem
Click on publish button next to submission or form. On displayed publish window choose one of servers JSON, REDCap or Ohmage JSON, leave all fields empty and click on publish button. You probably will have an impression that it succeeded. 

#### Expected behavior
It would be user-friendly to inform him what fields to fill/ disable publish button when parameters are not chosen/ some dialog message. We might need discussion here. 

#### Other information 
Things you tried, stack traces, related issues, suggestions on how to fix it...

ggalmazor:
Closed by #416

PR #416, opened 2019-02-12
Issue 210 validate publisher form

### LINKED ISSUE #208: User is asked for email address when publishing submissions/forms to servers.
Aggregate v1.4.x, Java v1.8.x, ...

#### Problem description
The user is asked for his "email address that will be granted access to these documents" when he is publishing data to JSON server,  Ohmage JSON Server and REDCap Server - but what documents? 

Click on publish button next to  submission or form and choose one of the servers and click on publish on the displayed window - you don't have to choose any url to test it (which is the other issue that I will report)

#### Expected behavior
User should be asked for email address only when he is publishing to Google Spreadsheet and Google Fusion Tables - here gaining access to documents does make a sense. 

#### Other information 
Screenshot with an example attached:
screenshot_exportred

### LINKED ISSUE #210: Too late error message when not choosing name workbook name when publishing to Google Spreadsheet.
Aggregate v1.4.x, Java v1.8.x, ...

#### Problem description
Too late error message when not choosing name workbook name when publishing to Google Spreadsheet.

Click on publish button next to submission or form. Choose to publish to Google Spreadsheet. Leave Workbook name field empty. Click on publish button. Dialog message asking you for email will appear. Input email address, confirm email address and finally, you will see error message informing you that "Spreadsheet name must be supplied."

#### Expected behavior
Error message shows which is good but it should display earlier or give users the chance to correct his action and input workbook name again or publish button may be disabled when user doesn't choose a workbook name.

#### Other information 
Screenshot of error message attached.
screenshot_error spreadsheet

ggalmazor:
Closes #208, #209, #210

Manually tried all field combinations in both publishers in the dev server.

No other approaches were considered. It's the narrowest change I could come up with that provides basic form validation.

An alternative would be to disable the button if the form is not valid. I don't have a strong opinion about this approach.

Common publisher form changes:
- Alerts will be shown when no type of publisher or kind of upload (all, new, existing) is selected
  There is really no way to test this without tampering with the form, although validating forms is always a good idea. These validations will be handy in case we change the form and we open the option of not selecting an option on those selects.

Changes to the Google Spreasheet publisher form:
- Users should expect a new publisher form that includes the owner's email:
  image
- Alerts will be shown if the user doesn't provide a workbook name or a valid owner's email address.

Changes to the JSON publisher form:
- It won't ask for the owner's email, since it's not really required to create it.
- In the publishers table, we use "N/A" in the "owner" column for JSON publishers.

Nope.

No.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -156,7 +156,7 @@ war {
 
 gretty {
   contextPath = '/'
-  host = 'localhost'
+  host = '192.168.1.134'
   logDir = 'logs'
 }
 
--- a/src/main/java/org/opendatakit/aggregate/client/popups/PublishPopup.java
+++ b/src/main/java/org/opendatakit/aggregate/client/popups/PublishPopup.java
@@ -249,7 +249,7 @@ public final class PublishPopup extends AbstractPopupBase {
 
           secureRequest(
               SecureGWT.getServicesAdminService(),
-              (rpc, sc, cb) -> rpc.createGoogleSpreadsheet(formId, workbookName, serviceOp, ownerEmail, cb),
+              (rpc, sc, cb) -> rpc.createGoogleSpreadsheet(formId, workbookName, serviceOp, "mailto:" + ownerEmail, cb),
               NO_OP_CONSUMER,
               this::onFailure
           );
--- a/src/main/java/org/opendatakit/aggregate/client/popups/PublishPopup.java
+++ b/src/main/java/org/opendatakit/aggregate/client/popups/PublishPopup.java
@@ -204,12 +204,14 @@ public final class PublishPopup extends AbstractPopupBase {
     public void onClick(ClickEvent event) {
 
       String externalServiceTypeString = serviceType.getSelectedValue();
-      ExternalServiceType type = (externalServiceTypeString == null) ? null :
-          ExternalServiceType.valueOf(externalServiceTypeString);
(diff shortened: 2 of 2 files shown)


kkrawczyk123:
@ggalmazor Validation on Google Spreadsheet works great! But I have noticed some issues:
- when I input email address I did not receive the email with access to sheet and in the published data tab emails are cut off, screenshot attached: 
screenshot from 2019-03-05 10-15-25

- Publish to JSON server is still allowed without input to any field on the dialog window. I think that URL field should be required one. 

ggalmazor:
Thanks, @kkrawczyk123!

I've added a couple of commits:
- One that fixes the issue with the cropped email addresses
- One that adds validation to the json publisher form (only checks the url field, and ensures that it's actually a URL)

ggalmazor:
I've left out url validation form the json publisher changes because it's more controversial than I thought it was going to be. There is no consensus on how a url should be validated in javascript (can't use Java for this one), and we will have to implement a custom regexp that meets all our needs: require http or https, support localhost, port numbers and simple paths.

In any case, if a user enters an invalid url, the publisher will fail and they will get feedback to make amends, therefore, I think it's safe to release this and improve it in a next release.

kkrawczyk123:
**Tested with success!**
Confirmed that issues have been fixed.

PR #421, opened 2019-02-12
Revert to previous code that had the webapp's path into account

### LINKED ISSUE #419: Fix Aggregate to work as non-ROOT webapp
Aggregate v2.0.0

#### Problem description
https://github.com/opendatakit/aggregate/pull/393/commits/7f2071bd96fb2cba31662bf2e88ba0422d5b4ee4 introduced a bug that will break the login page whenever Aggregate is not deployed as ROOT webapp ( at /)

Deploy Aggregate as a regular webapp and try to log in.

#### Expected behavior
It should log in

#### Other information 
Urls used in links and forms can't be truly absolute. They always need to be relative to the webapp's path in Tomcat.

ggalmazor:
Closes #419 

Run Aggregate in a local Tomcat both as ROOT and non-ROOT webapp, having configured a hostname and disabled the hostnames check.

In master, the login of the non-ROOT webapp won't work, and this PR solves that.

Also tested that this PR doesn't break the Cloud-Config stack by running it locally and verifying that the login still works.

This PR solves a regression bug introduced in v2.0.0. The change is to restore the original redirection code.

Nope.

No.

No.

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/servlet/ClearSessionThenLoginServlet.java
+++ b/src/main/java/org/opendatakit/aggregate/servlet/ClearSessionThenLoginServlet.java
@@ -63,11 +63,11 @@ public class ClearSessionThenLoginServlet extends ServletUtilBase {
     String newUrl;
     if (isAnon) {
       // anonymous user -- go to the login page...
-      newUrl = "/multimode_login.html";
+      newUrl = cc.getWebApplicationURL("multimode_login.html");
     } else {
       // we are logged in via token-based or basic or digest auth.
       // redirect to Spring's logout url...
-      newUrl = "/" + cc.getUserService().createLogoutURL();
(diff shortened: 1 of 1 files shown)


src/main/java/org/opendatakit/aggregate/servlet/ClearSessionThenLoginServlet.java
```diff
@@ -63,11 +63,11 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws
     String newUrl;
     if (isAnon) {
       // anonymous user -- go to the login page...
-      newUrl = "/multimode_login.html";
```
yanokwa: Did you mean to leave out the `/`?

kkrawczyk123:
**Tested with success!!**

PR #422, opened 2019-02-12
Issue 420 fix installer conf file assets

### LINKED ISSUE #420: Fix conf files in the installer
#### Problem description
The configuration files the installer deploys belong to the v1.x branch. `security.properties` and `odk-settings.xml` are outdated and need to be refreshed according to the example files at `/src/main/resources`

Run the installer, compare the files with the example files.

#### Expected behavior
They should have the same conf param keys

ggalmazor:
Closes #420, and #429

Build the installer, check the files it deploys, and deploy in local tomcat

Build the installer (linux-x64) and run it, then deploy the resulting WAR into a local Tomcat. Verify that Aggregate launches without changing anything.

The changes are conforming to any other task in the installer (set values in conf files, etc.)

This PR adds some required adaptations from v1.x to v2.x into the installer

In v1.x, the `odk-settings.xml` for each platform comes with a hard-coded driver classname, which deviates from the same file in the code repo, which uses whatever value is defined in the jdbc.driverClassName` props key in `jdbc.properties`

In v2.x, we have made an effort to normalize all the different configuration assets, and we changed the `odk-settings.xml` that the installer uses, but forgot to set the prop key in the jdbc settings file.

This PR takes care of that.

Nope, although we should test 

No.

No.

### DIFF
--- a/installer/project/components/configure_war.xml
+++ b/installer/project/components/configure_war.xml
@@ -254,6 +254,11 @@
         </actionGroup>
         <propertiesFileSet>
           <file>${temp_dir}/conf/common/security.properties</file>
+          <key>security.server.forceHttpsLinks</key>
+          <value>false</value>
+        </propertiesFileSet>
+        <propertiesFileSet>
+          <file>${temp_dir}/conf/common/security.properties</file>
           <key>security.server.port</key>
           <value>${http_port}</value>
         </propertiesFileSet>
@@ -269,11 +274,6 @@
         </propertiesFileSet>
         <propertiesFileSet>
           <file>${temp_dir}/conf/common/security.properties</file>
-          <key>security.server.superUser</key>
-          <value></value>
-        </propertiesFileSet>
-        <propertiesFileSet>
-          <file>${temp_dir}/conf/common/security.properties</file>
           <key>security.server.superUserUsername</key>
           <value>${username}</value>
         </propertiesFileSet>
@@ -284,9 +284,8 @@
         </propertiesFileSet>
         <propertiesFileSet>
           <file>${temp_dir}/conf/common/security.properties</file>
-          <key>security.help.about</key>
-          <value>
-            <![CDATA[auto-generated on ${creation.timestamp} for ${instance_display_name} ${product_fullname} ${product_version}]]></value>
+          <key>security.server.checkHostnames</key>
+          <value>true</value>
         </propertiesFileSet>
       </actionList>
     </actionGroup>
--- a/installer/project/components/configure_war.xml
+++ b/installer/project/components/configure_war.xml
@@ -101,67 +101,121 @@
       <actionList>
         <propertiesFileSet>
           <file>${temp_dir}/conf/${platform}/jdbc.properties</file>
-          <key>jdbc.help.about</key>
-          <value>Generated on ${creation.timestamp} for ${instance_display_name} ${product_fullname} ${product_version}</value>
+          <key>installer</key>
+          <value>Dummy key created by the installer</value>
         </propertiesFileSet>
       </actionList>
     </actionGroup>
     <actionGroup>
-      <!-- Sets the JDBC properties in jdbc.properties for MySQL and PostgreSQL -->
-      <explanation>Set jdbc properties</explanation>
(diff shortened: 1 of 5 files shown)


installer/project/components/configure_war.xml
```diff
@@ -284,9 +284,8 @@
         </propertiesFileSet>
         <propertiesFileSet>
           <file>${temp_dir}/conf/common/security.properties</file>
-          <key>security.help.about</key>
```
yanokwa: This removal seems dangerous for a point release. I realize it's set to an empty string in the various build processes we now use, but total removal of the key might have downstream effects that we aren't aware of.

installer/project/components/configure_war.xml
```diff
@@ -267,11 +272,6 @@
           <key>security.server.hostname</key>
           <value>${hostname}</value>
         </propertiesFileSet>
-        <propertiesFileSet>
```
yanokwa: See previous comment about removal of keys.

ripgrep shows `superUserEmail` is called by code in Aggregate, so it makes me nervous for a point release.

installer/project/components/params.xml
```diff
@@ -602,7 +602,7 @@ The installer will generate a script that can be run on the database server to c
           <description>Database Name:</description>
           <explanation></explanation>
           <value></value>
-          <default>odk_prod</default>
```
yanokwa: You are missing another `odk_prod` in installer/project/components/configure_war.xml. 

I catch these by always using [ripgrep](https://github.com/BurntSushi/ripgrep) when I'm making these sorts of changes.

installer/project/files/conf/postgresql/odk-settings.xml
```diff
@@ -57,7 +54,7 @@
   <bean id="user_service" class="org.opendatakit.common.security.spring.UserServiceImpl">
     <property name="datastore" ref="datastore"/>
     <property name="realm" ref="realm"/>
-    <property name="superUserEmail" value="${security.server.superUser}"/>
```
yanokwa: Don't forget this change if you agree with my previous comments.

installer/project/files/conf/mysql/odk-settings.xml
```diff
@@ -57,7 +54,7 @@
   <bean id="user_service" class="org.opendatakit.common.security.spring.UserServiceImpl">
     <property name="datastore" ref="datastore"/>
     <property name="realm" ref="realm"/>
-    <property name="superUserEmail" value="${security.server.superUser}"/>
```
yanokwa: Don't forget this change if you agree with my previous comments.

installer/project/components/configure_war.xml
ggalmazor: I think ripgrep is matching `security.server.superUserUsername` which is another different key. We don't use `security.server.superUser` since the removal of google account credentials.

installer/project/components/params.xml
ggalmazor: I think you were looking to a different commit... That line is already fixed

installer/project/components/configure_war.xml
ggalmazor: I'm worried about leaving it because I've seen prop files with a newline in the value. This prop key isn't used by any part of the code.

ggalmazor:
@yanokwa, after reviewing your suggestions, I'd want to stick with the changes I've pushed as they are. The installer should produce configuration files just like the ones at `src/main/resources/*.example`.

If you diff `src/main/resources/odk-settings.xml.example` with the three `odk-settings.xml` at `installer/project/files/conf/(mysql|postgresql|sqlserver)` you should get differences only in what regards to database artifacts. The `realm` bean should be the same, because Aggregate expects it to be like the one defined in the example file.

The same happens with the `security.properties` file generated by the installer. It must have exactly the same keys as the file at `security.properties.example`.

yanokwa:
@ggalmazor As you wish!

kkrawczyk123:
**Tested with success!**
I was able to generate war using the installer, deploy it and run aggregate. I've checked the content of aggregate settings files as well.

PR #432, opened 2019-03-04
Issue 429 Fix the JDBC conf file the installer creates

### LINKED ISSUE #429: Upgrade from 1.7.1 to 2.0.0 fails.
Linux (Docker), Aggregate v1.7.1 over tomcat:alpine (8.5.38-jre8), postgresql 10.7

#### Problem description
I upgrade from 1.5.0 to 1.6.1 to 1.7.0 to 1.7.1 without a problem but I can't upgrade to 2.0

#### Other information 
logs mention that problem : 
"Could not resolve placeholder 'jdbc.driverClassName' in string value "${jdbc.driverClassName}""

Thank's a lot.

ggalmazor:
Closes #429

This PR brings changes to the installer that ensure that all the jdbc properties are set by the installer.

Build the installer (linux-x64) and run it, then deploy the resulting WAR into a local Tomcat. Verify that Aggregate launches without changing anything.

This PR adds some required adaptations from v1.x to v2.x into the installer

In v1.x, the `odk-settings.xml` for each platform comes with a hard-coded driver classname, which deviates from the same file in the code repo, which uses whatever value is defined in the jdbc.driverClassName` props key in `jdbc.properties`

In v2.x, we have made an effort to normalize all the different configuration assets, and we changed the `odk-settings.xml` that the installer uses, but forgot to set the prop key in the jdbc settings file.

This PR takes care of that.

Nope. It only affects to the installer.

Nope.

Nope.

### DIFF
--- a/installer/project/components/configure_war.xml
+++ b/installer/project/components/configure_war.xml
@@ -101,67 +101,121 @@
       <actionList>
         <propertiesFileSet>
           <file>${temp_dir}/conf/${platform}/jdbc.properties</file>
-          <key>jdbc.help.about</key>
-          <value>Generated on ${creation.timestamp} for ${instance_display_name} ${product_fullname} ${product_version}</value>
+          <key>installer</key>
+          <value>Dummy key created by the installer</value>
         </propertiesFileSet>
       </actionList>
     </actionGroup>
     <actionGroup>
-      <!-- Sets the JDBC properties in jdbc.properties for MySQL and PostgreSQL -->
-      <explanation>Set jdbc properties</explanation>
-      <progressText>Setting jdbc properties</progressText>
+      <!-- Set PostgreSQL JDBC configuration -->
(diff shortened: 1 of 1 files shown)


ggalmazor:
@yanokwa I was thinking how to release this change, since it only affects to the installer. We will be releasing this fix on v2.0.1 for sure, but maybe we should backport it to v2.0 by replacing the installers. What do you think?

ggalmazor:
Installers:
- [Linux x64](https://drive.google.com/open?id=1MVy84AtELxjUuvbaz0kQQFHw1N6tGEF5)

yanokwa:
I don't like the idea of the same version numbers of the installers having different behavior. I think we should release in v2.0.1 and then we have a few options
* Pull the installers from the release page since we know theres a bug that prevents install
* Leave the installers as is

ggalmazor:
Closing in favour of #422, which has the commit from this PR. This will make testing the installer changes easier.

PR #439, opened 2019-03-07
Docker config improvements

brettneese:
_I'm working on deploying an Aggregate 2.0 instance to Kubernetes in the cleanest possible way, and documenting my process [here](https://www.brettneese.xyz/lab-notes-migrating-odk-aggregate-from-fargate-to-azure-kubernetes-service)._ 

This PR enables the ability to run the Gradle build commands for Docker without manually copying the `*.example` config files. I noticed that `docker.gradle` automatically overrides any settings in those files, anyway in the `dockerPrepare` build script so it made sense to have the build script just rename those files. Now you can build without manual work.

The only hacky thing I had to do was check the build commands sent to Gradle in the `checkConf` task against whether this was a build for Docker, which may be less than ideal, but seemed better than removing that check altogether. Otherwise if those files don't exist, Gradle throws an exception.

**I'm happy to entertain alternative approaches**, which is why I'm marking this WIP. I'm not very familiar with Gradle, so I'd love input there.

Part of the reason this matters is that the future I'm going to submit a PR that allows mounting those config files directly (instead of/in addition to the really hacky env var replacement thing going on in the entrypoint.) In K8S, this would be done using a ConfigMap -- you'll be able to just include the values of `security.properties` as a key in your ConfigMap, and mount it -- and I might be able to do something similar with Docker Compose. This seems a lot more sane that a mess of env vars that get replaced in files in the entrypoint.

`./gradlew clean dockerBuild -xtest -PwarMode=complete`

It may not be. In fact, it's quite hacky. I'd love to hear about other approaches. 

It may break other build processes. I haven't fully tested them, but it shouldn't.

Nope.

Yes, it will require an update to the Docker docs. I will be happy to update this PR with the appropriate changes if others agree this is a good idea.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Aggregate supports a variety of database engines, but we strongly recommend Post
 
 - Copy the `jdbc.properties.example`, `odk-settings.xml.example`, and `security.properties.example` files at `/src/main/resources` to the same location, removing the `.example` extension.
 
-  If you have followed the database configuration steps above, you don't need to make any change in these files. Otherwise, head to the [Aggregate configuration guide](docs/aggregate-config.md) and make the required changes for your environment.
+  If you have followed the database configuration steps above, you don't need to make any change in these files. Otherwise, head to the [Aggregate configuration guide](docs/aggregate-config.md) and make the required changes for your environment. Note that if you are running in Docker this step is not necessary as the build scripts copy those files for you and automatically override any values set in those files. See [here](docs/build-and-run-a-docker-image.md)].
   
 - Start a local development Aggregate server with `./gradlew appRunWar`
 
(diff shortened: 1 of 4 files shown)


ggalmazor:
Hi, @brettneese!

This PR is awesome, thanks!!! I think we can move forward with that `if` in the `checkConf` task, no problem.

If you throw in the docs changes, I think we can approve the current commits as they are.

I'm very much looking forward to those K8S instructions too. This is exciting!

brettneese:
@ggalmazor Excellent! I did make one small change, and that was to remove the warning. If the files get overridden, then there's no point in warning anyway that they're missing (if anything, we should warn people that they're being overriden! But I'm noting that in the docs.)

brettneese:
And I updated the docs!

ggalmazor:
Since you tagged this PR as WIP, I'll wait for your confirmation in order to merge this PR

brettneese:
I just marked it as not WIP, it's good to go! 👍  @ggalmazor

PR #441, opened 2019-03-11
Issue 440 Fix change password url generation

### LINKED ISSUE #440: Change passwords doesn't work when deployed as non-ROOT webapp
Aggregate v2.0.0

#### Problem description
[Forum post](https://forum.opendatakit.org/t/json-change-password-request-to-ssl-user-manage-passwords-failed/18353/13)

- Deploy Aggregate as a non-ROOT webapp in Tomcat
- Try to change the password of a user

Currently, the call to the web API should get an HTTP 404 error 

#### Expected behavior
Changing the password should work

#### Other information 
[This commit](https://github.com/opendatakit/aggregate/commit/048410e8ba18900e642d18db6aa60acf9fa184e2) solved the same issue for the login process, but there are still other places in the code that should be fixed:
- `SecurityServiceImpl.getRealmInfo()` method

ggalmazor:
Closes #440 

Launches as non-ROOT webapp in a local tomcat server and changed the password while inspecting the network tab using Chrome's developer console.

This fixes the issue like in https://github.com/opendatakit/aggregate/commit/048410e8ba18900e642d18db6aa60acf9fa184e2

Nope.

No.

No.

### DIFF
--- a/src/main/java/org/opendatakit/common/security/server/SecurityServiceImpl.java
+++ b/src/main/java/org/opendatakit/common/security/server/SecurityServiceImpl.java
@@ -111,7 +111,7 @@ public class SecurityServiceImpl extends RemoteServiceServlet implements
       throw new DatastoreFailureException("Unable to access datastore");
     }
     // User interface layer uses this URL to submit password changes securely
-    r.setChangeUserPasswordURL("/" + UserManagePasswordsServlet.ADDR);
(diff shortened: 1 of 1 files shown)


zwets:
I just tested this by replacing only the `SecurityServiceImpl.class` in my deployed war and can confirm the issue is solved here.

kkrawczyk123:
I've tested changing superuser and user passwords, everything worked great for me!

PR #445, opened 2019-03-18
WIP - Support for other cloud providers

ggalmazor:
This PR adds variations on the Digital Ocean Cloud Config stack to use it in other cloud providers:
- [x] AWS EC2 (based on Digital Ocean's Cloud Config stack)
- [x] Google Compute Engine VM (startup shell script based on the Cloud Config stack)
  GCE doesn't support `runCmd` blocks in Cloud Config
- [x] Azure


cloud-config/aws/README.md
```diff
@@ -0,0 +1,156 @@
+This Cloud-Config setup is intended to be used with the Ubuntu Server 18.04 LTS AMI in AWS. You can search for the latest AMI id for your AWS zone at https://cloud-images.ubuntu.com/locator/ec2. Search for `18.04 amd64 hvm:ebs-ssd`
```
yanokwa: I think you can drop the bit about latest AMI ID. I don't think anyone should care.

cloud-config/aws/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+**To use this setup, you must able to link a domain name to EC2 machine's IP address. If you don't own a domain, services such as [FreeDNS](https://freedns.afraid.org) offer free sub-domains under a range of domains.**
+
+## Prerequisites
+
+**Warning**: Make sure at all time that you have selected the availability zone where you want to perform your actions. You can choose the availability zone using the dropdown menu at the top-right corner of the AWS console website.
+
+### Create a VPC
```
yanokwa: I would expect most people would use the default VPC. If this is best practice, or makes other things harder, maybe it's not too bad.

cloud-config/aws/README.md
```diff
@@ -0,0 +1,156 @@
...
+  
+  | Type | Source |
+  | --- | --- |
+  | SSH | Anywhere |
+  | HTTP | Anywhere |
+  | HTTPS | Anywhere |
+  
+  Click on `Save rules`.
+
+## Create an IAM role
```
yanokwa: Why is this role needed? Is it to grab the tag?

cloud-config/aws/README.md
```diff
@@ -0,0 +1,156 @@
...
+- Give the new IAM role a name
+
+  Click on `Create role`.
+
+## Create the EC2 machine
+
+- Go to the [EC2 Dashboard](https://console.aws.amazon.com/ec2/v2/home#Home:)
+  
+  Click on `Launch instance`
+- Search for the `Ubuntu Server 18.04` AMI
```
yanokwa: If you search for `Ubuntu Server 18.04 LTS ` then you'll only get one option, which is what we want.

cloud-config/aws/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+  You should select the AMI showing `Ubuntu Server 18.04 LTS (HVM), SSD Volume Type` tag line.
+  
+  Make sure you select the `64-bit (x86)` option under the `Select` button.
+  
+  Click on `Select`.
+- Select the instance type you want to use.
```
yanokwa: I think we should recommend a minimum size and save people the thinking.

cloud-config/aws/README.md
```diff
@@ -0,0 +1,156 @@
...
+  
+  Select `Enable` in the `Auto-assign Public IP` dropdown.
+  
+  Select the IAM role you've created in the previous step from the `IAM role` dropdown.
+  
+  Toggle the `Advanced Details` section.
+  
+  Copy the contents of the [Aggregate Cloud Config stack for AWS](https://raw.githubusercontent.com/opendatakit/aggregate/master/cloud-config/aws/cloud-config.yml) into the `User data` text box.
+
+  Click on `Next: Add Storage`.
```
yanokwa: If we pick a reasonable size machine (e.g.,  match the machine size on DO?) then this step goes away.

yanokwa:
@ggalmazor I think we should skip Linode until someone complains. We've had requests for Azure and GCE.

cloud-config/google-cloud/startup-script.sh
```diff
@@ -0,0 +1,89 @@
...
+ zip \
+ unzip \
+ wget \
+ curl \
+ tomcat8 \
+ postgresql-10 \
+ openjdk-8-jdk-headless \
+ nginx \
+ software-properties-common \
+ python-certbot-nginx
```
yanokwa: This step failed for me trying to configure tzdata, which I think comes with certbot. Log at http://dpaste.com/1CCQFWQ.

According to https://askubuntu.com/questions/909277, it might work if you sent the `DEBIAN_FRONTEND` variable to `noninteractive`. https://stackoverflow.com/questions/8671308/non-interactive-method-for-dpkg-reconfigure-tzdata might also help.

cloud-config/google-cloud/README.md
```diff
@@ -0,0 +1,62 @@
+**To use this setup, you must be able create DNS domain entries. If you don't own a domain, services such as [FreeDNS](https://freedns.afraid.org) offer free sub-domains under a range of domains.**
+
+## Create the GCE VM
```
yanokwa: I think it'd be nice if we give suggestions on location (close to where the data entry) and machine size.

cloud-config/aws/README.md
ggalmazor: I believe you can't do that anymore... I'll check

ggalmazor: Yeah, unless theres some hidden way I don't know about, I've confirmed you can't complete the creation of the EC2 machine without providing a VPC.

cloud-config/aws/README.md
ggalmazor: Do you have a figure in mind? The default is 8 GiB, which feels too few.

cloud-config/aws/README.md
ggalmazor: Most of the instance types are EBS only storage (especially t1, including the free tier type), which requires to add a storage volume. 

The instance types that include storage are kind of expensive, but maybe we should add something like "If you select an instance type with built-in storage, you can skip this section."

cloud-config/aws/README.md
yanokwa: I'd try to match what Digital Ocean provides. So maybe a t2.small (1 vCPUs, 2 GB RAM) with 30 GB of RAM is the absolute minimum for small deployments. I haven't tried it, but it should be reasonable.

cloud-config/aws/README.md
yanokwa: You can customize the EBS volumes with bigger root drives, no? Does that have some downside I'm not thinking of?

cloud-config/azure/README.md
```diff
@@ -0,0 +1,89 @@
+https://portal.azure.com/#blade/HubsExtension/Resources/resourceType/Microsoft.Compute%2FVirtualMachines
```
yanokwa: Extra link?

cloud-config/azure/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+**To use this setup, you must able to link a domain name to EC2 machine's IP address. If you don't own a domain, services such as [FreeDNS](https://freedns.afraid.org) offer free sub-domains under a range of domains.**
+
+## Prerequisites
+
+**Warning**: Make sure at all time that you have selected the availability zone where you want to perform your actions. You can choose the availability zone using the dropdown menu at the top-right corner of the AWS console website.
```
yanokwa: Warning needs to be customized to Azure.

cloud-config/azure/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+  Fill in the credentials of the user account that will be created during the creation of the VM. You will use this user account to complete the Aggregate installation later.
+  
+- Review the Inbound Port Rules section.
+
+  Select the `Allow selected ports` option in the `Public inbound ports` field.
+  
+  Select `HTTP`, `HTTPS`, and `SSH` in the `Select inbound ports` field.
+  
+- Review the Network configuration.
```
yanokwa: The more natural thing is to click to at the bottom to get to the disks, so point out the Networking tab at top (or tell em to click through disks).

cloud-config/azure/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+  Click on the `Networking` tab.
+  
+  Click on the `Create new` link under the `Public IP` field dropdown, give it a name, and click `OK`.
```
yanokwa: Mine was already named ($MyServerName-ip).

cloud-config/azure/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+- Review the Guest Config settings.
+
+  Click on the `Guest Config` tab.
```
yanokwa: I don't see a Guest Config tag, but I do see an Advanced tab.

cloud-config/azure/README.md
```diff
@@ -0,0 +1,89 @@
...
+  
+- Review the tags.
+
+  Click on the `Tags` tab.
+  
+  Add a tag with the `aggregate.hostname` name, and write the DNS domain name you want to use as the value.
+  
+- Launch the VM
+
+  Click on the `Review + Launch` tab.
```
yanokwa: Mine says Review + Create

cloud-config/azure/README.md
```diff
@@ -0,0 +1,89 @@
...
+  
+  Click on `Launch`.
+
+## Set up your domain
+
+1. Take note of the public IP address of your VM (e.g., 12.34.56.79) and set a *DNS A record* pointing to it.
+
+  You can get your instance's IP address on the [Azure Virtual Machines Dashboard](https://portal.azure.com/?l=en.en-us#blade/HubsExtension/Resources/resourceType/Microsoft.Compute%2FVirtualMachines). After clicking on the VM from the list, check the value of the `Public IP address` field.
```
yanokwa: Give me a hint that it's in the overview.

cloud-config/azure/cloud-config.yml
```diff
@@ -0,0 +1,94 @@
...
+  - su postgres -c "psql -c \"CREATE ROLE aggregate WITH LOGIN PASSWORD 'aggregate'\""
+  - su postgres -c "psql -c \"CREATE DATABASE aggregate WITH OWNER aggregate\""
+  - su postgres -c "psql -c \"GRANT ALL PRIVILEGES ON DATABASE aggregate TO aggregate\""
+  - su postgres -c "psql -c \"CREATE SCHEMA aggregate\" aggregate"
+  - su postgres -c "psql -c \"ALTER SCHEMA aggregate OWNER TO aggregate\" aggregate"
+  - su postgres -c "psql -c \"GRANT ALL PRIVILEGES ON SCHEMA aggregate TO aggregate\" aggregate"
+
+  - curl -H Metadata:true "http://169.254.169.254/metadata/instance/compute/tags?api-version=2017-08-01&format=text" | awk -F':' '{print $2}' > /tmp/domain-name
+  - sed -i -e 's/foo\.bar/'"$(cat /tmp/domain-name)"'/' /root/aggregate-config.json
+  - sed -i -e 's/foo\.bar/'"$(cat /tmp/domain-name)"'/' /etc/nginx/sites-enabled/aggregate
```
yanokwa: The previous sed command worked, but this one seems to have failed. When I tried to run it as non-root user, I got this error. 
```
$ sed -i -e 's/foo\.bar/'"$(cat /tmp/domain-name)"'/' /etc/nginx/sites-enabled/aggregate
sed: couldn't open temporary file /etc/nginx/sites-enabled/sedmuulct: Permission denied
```
Running it as root worked.

cloud-config/aws/README.md
ggalmazor: EBS volumes are fully customizable, and you get to have snapshots and even scale your instance up or down without having to reinstall software or migrate data.

I've seen how AWS has been progressively moving away from built-in storage to EBS only storage, at least on basic and general purpose instance types.

For specific high IO applications, there are still instance types that come with built-in specialized storage.

cloud-config/google-cloud/startup-script.sh
ggalmazor: How did you get the logs of the startup script?

ggalmazor: syslog! nevermind :)

cloud-config/azure/README.md
ggalmazor: Honestly, I get something different each time, but so far, I've always had something that would ensure there's a public IP pre-filled there, so I'm skipping the section all together.

ggalmazor: Also, I can't believe that I forgot to add a disk to the instructions!

cloud-config/azure/cloud-config.yml
ggalmazor: I haven't been able to reproduce this issue... How did you notice this had failed?

ggalmazor:
I think I've gone through all your feedback, @yanokwa. Thanks!

PR #423, opened 2019-02-13
fix bad caching headers

terencehonles:
Fixes: https://forum.opendatakit.org/t/invalid-character-in-last-modified-header-tomcat8-with-odk-aggregate/9162

#### What has been done to verify that this works as intended?
Currently cherry-picking the change to the release we are using and building a new jar.

#### Why is this the best possible solution? Were any other approaches considered?
Bug fix

#### Are there any risks to merging this code? If so, what are they?
None

#### Do we need any specific form for testing your changes? If so, please attach one
No

#### Does this change require updates to documentation? If so, please file an issue at https://github.com/opendatakit/docs/issues/new and include the link below.
No

PR #448, opened 2019-03-26
fix bad caching headers

terencehonles:
Backport of #423 to v1.X

cc: @ggalmazor 

We have been running this in production since #423 was opened


terencehonles:
Sounds good to me!

PR #451, opened 2019-04-02
URL validation added

### LINKED ISSUE #436: Validate JSON publisher url
#416 introduces validation on the publisher creation forms.

The JSON publisher form only checks that the user enters something in the url field. We should also verify that it's a valid URL string.

sd1998:
Closes #436 

Tested the regular expression across multiple URLs.

No


sd1998:
@ggalmazor I have fixed the issue.

sd1998:
@ggalmazor I have made the changes. Here are the screenshots of the test I have performed:-
image
image

ggalmazor:
The extra urls I've tried are:
- http://192.168.1.129:8080
- http://foo.bar/foo/bar?foo=bar

src/main/java/org/opendatakit/aggregate/client/popups/PublishPopup.java
```diff
@@ -294,4 +295,8 @@ public native static boolean validateEmail(String email) /*-{
     return re.test(String(email).toLowerCase());
   }-*/;
 
+  public boolean validateURL(String url){
+    return url.matches("(http?|https?)://[-a-zA-Z0-9+&@#/%?=~_|!:,.;]*[-a-zA-Z0-9+&@#/%=~_|]");
```
yanokwa: This feels fragile to me. I think we either do something very naive (just check for `(http?|https?)://`) or we use the internal Java library (or something from Apache if it's already embedded).

ggalmazor: I thought about using `URL` for that, but we can't use it because there's no GWT translation of `URL`. 

Unfortunately, there's no built-in URL validation in JavaScript, so people use regexps like the one @sd1998 is proposing (there are plenty of custom validators out there).

The js validation library [validate.js](https://validatejs.org/#validators-url) uses [this regexp to validate URLs](https://gist.github.com/dperini/729294). we could use that one instead.

kkrawczyk123:
I’ve checked empty value, trailing spaces, extra characters, and valid urls too. I have noticed that I am able to use “http://localhost” without adding the port number or even urls like “https://bla” which are not exactly valid ones (like on attached screenshot)
Screenshot from 2019-04-09 13-13-32

PR #456, opened 2019-04-11
Issue 452 The installer is not configuring the hostname correctly

### LINKED ISSUE #452: The installer is not configuring the hostname correctly
#### Problem description
It looks like the installer is leaving the `security.server.hostname` field in the `security.properties` conf file empty regardless of what the user enters in the SSL configuration installer page.

This shouldn't be a problem unless Aggregate doesn't automatically detect the correct hostname during the start process.

Use the installer and configure SSL.

#### Expected behavior
The installer should fill the conf files correctly and set a value to the `security.server.hostname` field in the `security.properties` conf file.

ggalmazor:
Closes #452

Ensure that the installer sets the hostname in output conf files
- Previously, we had a post-installation action that would copy the value of raw_hostname to hostname and perform transformations conditionally if the user was using GAE
- When we removed support for GAE, we broke this by removing all the post-installation actions related to the hostname variable
- In this commit, raw_hostname becomes hostname, fixing the issue
- This commit also adds a couple of hints to help the user fill in a value.
- Users are reminded that the hostname must match their SSL certificate.

Screenshot of the `Apache Tomcat Port Configuration` screen when `no SSL` is selected:
issue 451 screenshot 2

Screenshot of the `Apache Tomcat Port Configuration` screen when `SSL` is selected:
issue 452 screenshot 1

Build the installer, run it and verify that the output `security.properties` is correctly configured with whatever value is set in the `Internet-visible IP address or DNS name` field

This fixes a bug introduced when removing GAE support from the current master branch

None.

No.

No.

### DIFF
--- a/installer/project/components/params.xml
+++ b/installer/project/components/params.xml
@@ -21,17 +21,6 @@
       <cliOptionText>Output a .options file that contains the configuration from running the installer.</cliOptionText>
     </booleanParameter>
     <stringParameter>
-      <name>hostname</name>
-      <description>Hidden value used to hold resolved fully-qualfied hostname.</description>
-      <explanation></explanation>
-      <value></value>
-      <default></default>
-      <allowEmptyValue>1</allowEmptyValue>
-      <ask>0</ask>
-      <insertBefore>${parent_installdir}</insertBefore>
-      <width>40</width>
-    </stringParameter>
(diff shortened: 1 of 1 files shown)


kkrawczyk123:
**Tested with success!**

PR #458, opened 2019-04-11
Issue 457 relax json publisher url validation

### LINKED ISSUE #457: Prevent false positives when validating a JSON publisher URL
#### Problem description
#451 introduced URL validation into the JSON publisher creation dialog using a custom regular expression.

We can't really be sure that the regexp won't produce false positives and reject valid URLs. 

We want to be on the safe side and prevent giving users a hard time unnecessarily. Instead of complete URL validation, we would want to relax the URL validation to a bare minimum that guarantees that we won't even produce false positives.

Checking that URL starts with `http://` or `https://` could be enough for us at this moment.

ggalmazor:
Closes #457 

Run Aggregate locally and try out the new URL validation with URLS that start with `http://`, `https://` or otherwise.

This is a super narrow change. No alternatives where considered.

Nope.

No.

No.


kkrawczyk123:
**Tested with success!**
Verified on tomcat instance.

PR #460, opened 2019-04-22
Issue 459 avoid page load blocks

### LINKED ISSUE #459: Aggregate should not block page loads when network is down
Aggregate v2.0.2

#### Problem description
If you run Aggregate with the network down, the front page does not load. The two network calls that I can see are
* For maps.googleapis.com/maps/api/js in AggregateHTMLServlet.java 
* For api.github.com in LayoutUtils.java

#### Potential solutions
* Add async and defer to Google maps call as shown in the docs at https://developers.google.com/maps/documentation/javascript/tutorial
* If the XMLHttpRequest call fails, return a versionNote of "Version check failed" rather than blocking the UI.

ggalmazor:
Closes #459

I've used Chrome's inspector for all these tests with the "Disable cache" from the network tab enabled.

**Version note blocking issues**
- Start Aggregate in `localhost` and verify that I get the "you're up to date" note. 
- Restart Aggregate, disable the network connection and verify that I get the "version check failed" note.

**Google maps JS blocking issues**
- Start Aggregate and verify with the inspector that the script gets eventually loaded.
- Restart Aggregate, disable the network connection and verify that I get an error but the page still loads.

First I tried to modify the client's JSNI implementation in `LayoutUtils.java` to handle errors related to the requests to GitHub's API, but I got to very complicated half-solutions due to conflicts between the blocking nature of GWT Java UI building and the asynchronous XMLHttpRequest.

Then I decided to switch to an all-Java server-side solution. At the expense of having a new GWT service, now we completely avoid client-side JSNI code and the asynchronous interaction is entirely made through GWT RPC calls, which is results in much simpler client code.

Thanks to moving this code to the server, we're also able to throttle-control calls to GitHub's API (and avoid nasty temporary bans due to spamming it with requests). Now, instead the clients (potentially many) doing calls to GH, it's the server who does 1 request per 24 hours max.

None.

No.

No.

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/client/LayoutUtils.java
+++ b/src/main/java/org/opendatakit/aggregate/client/LayoutUtils.java
@@ -3,28 +3,14 @@ package org.opendatakit.aggregate.client;
 import static org.opendatakit.aggregate.buildconfig.BuildConfig.VERSION;
 
 import com.google.gwt.dom.client.Style;
+import com.google.gwt.user.client.rpc.AsyncCallback;
 import com.google.gwt.user.client.ui.HTML;
 import com.google.gwt.user.client.ui.UIObject;
 
 public class LayoutUtils {
-  private static String latestVersion;
-
-  static native String getLatestVersion() /*-{
-    var req = new XMLHttpRequest();
-    req.open('GET', 'https://api.github.com/repos/opendatakit/aggregate/releases/latest', false);
-    req.send(null);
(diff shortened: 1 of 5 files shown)


lognaturel:
Moving the version check to the server seems great all around. 👍 Would be good to get a QA sanity check and ideally it can go in 2.0.3.

kkrawczyk123:
**Tested with success!**

PR #469, opened 2019-05-20
allow any 2XX status code from an external JSON server

terencehonles:
When a server responds with another 2XX code, such as 201 the server treats this as an error. This accepts any 2XX response.

I used the number 300 instead of SC_MULTIPLE_CHOICES as that constant's name doesn't signify the start of the non success range.

We'll be deploying this shortly.

Any 2XX code is a success code.

No.

No

### DIFF
--- a/src/main/java/org/opendatakit/aggregate/externalservice/JsonServer.java
+++ b/src/main/java/org/opendatakit/aggregate/externalservice/JsonServer.java
@@ -134,7 +134,7 @@ public class JsonServer extends AbstractExternalService implements ExternalServi
       }
       if (statusCode == HttpServletResponse.SC_UNAUTHORIZED) {
         throw new ODKExternalServiceCredentialsException(reason + " (" + statusCode + ")");
-      } else if (statusCode != HttpServletResponse.SC_OK) {
+      } else if (statusCode < HttpServletResponse.SC_OK || statusCode >= 300) {
         throw new ODKExternalServiceException(reason + " (" + statusCode + ")");
       }
     } catch (ODKExternalServiceException e) {


ggalmazor:
I think this is safe to be merged directly. I don't know how we could test this manually anyway. Do you agree, @lognaturel?

yanokwa:
I'm not the world's greatest API designer or anything, but seems like we should be narrower here and only allow 200, 201, and 202. I don't feel strongly about it...

terencehonles:
From https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml

> 2xx: Success - The action was successfully received, understood, and accepted
> 
> 200 | OK | [RFC7231, Section 6.3.1]
> 201 | Created | [RFC7231, Section 6.3.2]
> 202 | Accepted | [RFC7231, Section 6.3.3]
> 203 | Non-Authoritative Information | [RFC7231, Section 6.3.4]
> 204 | No Content | [RFC7231, Section 6.3.5]
> 205 | Reset Content | [RFC7231, Section 6.3.6]
> 206 | Partial Content | [RFC7233, Section 4.1]
> 207 | Multi-Status | [RFC4918]
> 208 | Already Reported | [RFC5842]
> 209-225 | Unassigned |  
> 226 | IM Used | [RFC3229]
> 227-299 | Unassigned

200, 201, 202, 204, 207, and 208 should be acceptable, but according to the document all of the 200 range is a "success" and there probably is not a strong reason to reject the couple that fall through and the unassigned ranges just because they don't "seem" to make sense. If the server received, understood, and accepted the request that's pretty much all the ODK server cares about.

ggalmazor:
I agree with @terencehonles. I think it's safe to assume any incoming 2xx response completes successfully the request.

PR #470, opened 2019-05-20
allow any 2XX status code from an external JSON server

terencehonles:
Backports #469 


lognaturel:
@ggalmazor is there a reason that this wasn't merged? It seems it should be in v1.7.4.

PR #475, opened 2019-07-23
Add an error resolution

soloincc:
Provide a workaround when running gradle for the first time

Closes #


ggalmazor:
Hi, @soloincc!

First, let me apologize for the delay of this review.

I think the current instructions give enough information to work around the error you're referring to:
- This errors is caused by cloning the repo without having installed git LFS first
- In any case, we don't want devs to use Gradle directly in our projects. Instead, we want them to use the `gradlew` wrapper script that is already present in the repo.

Both things are already explained in the docs. Do you feel we should expand on any of those? If not, I'd close this PR without merging.

PR #486, opened 2020-04-20
Added -L to curl to account for github API redirects

geom-wgr1:
Very small change because GitHub started redirecting its API calls. I've confirmed this works while setting up a test Aggregate server on Azure today.


lognaturel:
Thanks so much for getting right on that, @geom-wgr1! Closing in favor of https://github.com/getodk/aggregate/pull/487 to make sure the change is made everywhere needed.

Really appreciate you doing this and make sure you say hello [on the community forum](https://forum.getodk.org/) if you haven't already!

PR #487, opened 2020-04-22
Instruct curl to follow redirects

yanokwa:
We have renamed opendatakit/aggregate to getodk/aggregate. Web browsers follow the redirect by default, but curl doesn't unless you add `-L`.


lognaturel:
Merging now even though the build is broken to get the fix out as soon as possible. @yanokwa will be fixing the build shortly.

PR #488, opened 2020-04-22
Fix build

yanokwa:
Tested by SSHing into a build machine and trying it out.


lognaturel (approved):

lognaturel:
Looks like this works for mysql but not for postgres. Not sure why the postgres build didn't run (that's why the commit gets a green check).

PR #492, opened 2020-05-25
Update slug

yanokwa:


lognaturel:
Depends on https://github.com/getodk/javarosa/pull/577 which is currently marked as draft.

